PR #8, opened 2022-12-30
Refactor graphserver

### LINKED ISSUE #4: Attaching to running graphs fails sporadically
To reproduce:
* Run `ezmsg_toy.py`
* Run `ezmsg_attach.py` in another terminal
    * Sometimes `ezmsg_attach.py` only publishes one message and hangs
* Interrupt `ezmsg_attach.py` with ^c
* Run `ezmsg_attach.py` again
    * Usually `ezmsg_attach.py` only reports one message and hangs
* Interrupt `ezmsg_attach.py` with ^c
    * This interrupt fails and the terminal must be killed because `sigint` is captured
* Run `ezmsg_attach.py` again
    * This causes both `ezmsg_toy.py` and `ezmsg_attach.py` to crash

griffinmilsap:
* Significantly reduces technical debt in `GraphServer`, `Publisher` and `Subscriber`
* Fixes #4 
* Adds a test for attaching to running graphs
* Adds a `__version__` attribute and configures packaging to use that attribute for version
* Version bump to `3.1.2`
* Exposes `GraphServer` and `GraphContext` at the top level of `ezmsg.core`
* Strong version checking between connecting clients and `GraphServer`


pperanich (approved):
Everything looks good to me!

PR #9, opened 2023-01-06
Single process

griffinmilsap:
The pull request rounds out necessary changes for `3.1.2` including:

* **Single process mode**: If all tasks would run in one process, and a `GraphServer`/`SHMServer` combo are already up, `ez.run` will not attempt multiprocessing.  Critical for using `ezmsg` in REPL/notebook environments.
* **Associate `Publisher` and `Subscriber` objects with `OutputStream` and `InputStream` instead of individual tasks**: 
    * Should drastically reduce the number of connections/sockets required for systems, enhancing performance
    * Allows for control over individual `OutputStream`... 
        * _hostname/port specification;_ allowing outbound TCP communications to other PCs
        * _number of buffers;_ down to 1 buffer for direct control of channel latency and backpressure
        * _initial SHM allocation size;_ to avoid resizing SHM while running
        * _force TCP message passing;_ for small message sizes, there's no performance loss and this allows us to run perf benchmarks including TCP transmission without multi machine CI setups.
    * **IMPORTANT BREAKING CHANGE** -- it is no-longer possible to subscribe to `OutputStream`s and publish to `InputStream`s.  This was previously possible (unintended) and should be fixed in downstream projects.  If you can come up with a valid use-case, please let me know.
* **Command line tool** that allows for serving/shutting down `GraphServer` and `SHMServer` from command line.  After installing `ezmsg`, simply run `ezmsg serve` from the command line to start up the `ezmsg` background servers

This pull request also:
* Fixes several small bugs
* Switches `SHMServer` and `GraphServer` ports (25978/9).  GraphServer now sits at 25978 by default, and SHMServer now sits at 25979 by default. 
    * Incidentially, this was driven by trying to run the servers through SSH forwarding on a remote machine.  I learned we should NEVER forward the `SHMServer` port (oops; seems pretty obvious to me now).  Figured we should set the GraphServer to the base port -> 25978 represented as a 16-bit unsigned number in binary is 0110010101111010, which when represented by two ascii characters is the string `ez` 👍  
* `SHMServer`, `GraphServer` and the publisher port-range start are now individually configurable by environment variables.  
* `GraphServer`'s `pause` and `resume` are no-longer asynchronous; don't need to be `await`'ed
* Shuts down immediately after `NormalTermination` is raised, rather than waiting up to 0.5 seconds
    * Tests run ever-so-slightly-faster because of this


griffinmilsap:
Unsure if these feature additions merit a minor version bump (`3.2.0`) instead of a bugfix bump (`3.1.2`); would welcome feedback on that.

ezmsg/core/graphserver.py
```diff
@@ -139,104 +127,93 @@ async def api(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter)
...
-                    writer.write(Command.COMPLETE.value)
-                    await writer.drain()
+                    elif req == Command.RESUME.value:
+                        for pub in self._publishers():
+                            pub.writer.write(Command.RESUME.value)
 
-                else:
-                    logger.warn(f'GraphConnection API received unknown command {req}')
+                    else:
+                        logger.warn(f'GraphSerer received unknown command {req}')
```
pperanich, line 216: `GraphSerer` -> `GraphServer`

pperanich (changes requested):
Looks good besides one log typo!

PR #37, opened 2023-10-06
Add files via upload

itsJorgeRivera:
Adding a Spectral Extractor that uses the CCA algorithm for SSVEP applications. In settings, we define a number of Frequencies of Interest that we want to survey for in an incoming data stream. The SpectraCcaExtractor unit then checks the incoming stream for those frequencies of interest and returns the strongest one.


griffinmilsap (commented):
I'm thinking this is actually a better contribution for ezmsg-ssvep -- this implementation is actually pretty SSVEP specific.  @itsJorgeRivera could you make a PR [here](https://github.com/griffinmilsap/ezmsg-ssvep) instead?

PR #42, opened 2023-10-26
Fixes synth.Clock sleep isues

### LINKED ISSUE #41: sigproc.synth.Clock sleeps too long
https://github.com/iscoe/ezmsg/blob/005590f74ec66b22cd36a94827ff5822ee4bbc92/extensions/ezmsg-sigproc/ezmsg/sigproc/synth.py#L42

While this loop is very tight, the sleep time fails to account for what might be happening in the other asynchronous workloads. When running this as is, I was only getting about 990 samples per second when I specified the rate as 1 kHz.

cboulay:
Fixes #41 

With this change, at a nominal rate of 1 kHz, instead of pushing 990 samples/sec I am now pushing around 999.8. My smoothed running average sometimes reports above 1000 but it's usually in the 999.6-999.8 range. I don't know why it isn't bang-on 1000.0 but it could be a precision error in my running average.

I also snuck in an addition to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# JetBrains
+.idea/
--- a/extensions/ezmsg-sigproc/ezmsg/sigproc/synth.py
+++ b/extensions/ezmsg-sigproc/ezmsg/sigproc/synth.py
@@ -37,9 +37,13 @@ class Clock(ez.Unit):
 
     @ez.publisher(OUTPUT_CLOCK)
     async def generate(self) -> AsyncGenerator:
+        t_0 = time.time()
+        n_dispatch = 0
         while True:
             if self.STATE.cur_settings.dispatch_rate is not None:
-                await asyncio.sleep(1.0 / self.STATE.cur_settings.dispatch_rate)
+                n_dispatch += 1
+                t_next = t_0 + n_dispatch / self.STATE.cur_settings.dispatch_rate
+                await asyncio.sleep(t_next - time.time())
             yield self.OUTPUT_CLOCK, ez.Flag
 
 


.gitignore
```diff
@@ -138,3 +138,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# JetBrains
+.idea/
```
griffinmilsap, line 143: I like to add stuff like this to my [global .gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

extensions/ezmsg-sigproc/ezmsg/sigproc/synth.py
```diff
@@ -37,9 +37,13 @@ async def on_settings(self, msg: ClockSettings) -> None:
 
     @ez.publisher(OUTPUT_CLOCK)
     async def generate(self) -> AsyncGenerator:
+        t_0 = time.time()
+        n_dispatch = 0
         while True:
             if self.STATE.cur_settings.dispatch_rate is not None:
-                await asyncio.sleep(1.0 / self.STATE.cur_settings.dispatch_rate)
+                n_dispatch += 1
```
griffinmilsap, line 44: Although this will suffer from floating point accuracy issues if the code runs long enough, this is definitely more accurate than what I had before.  If folks complain down the road, we could add an occasional reset of t_0 and n_dispatch to keep them below some threshold to avoid floating point inaccuracies.

griffinmilsap (approved):
Not going to be pedantic and request the removal of a single line in the .gitignore here haha...  Happy to have your contribution :)

.gitignore
cboulay: Fair enough. If you ask GitHub to create a Python .gitignore for you, it gives you this block:

```
# PyCharm
#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
#  and can be added to the global gitignore or merged into this file.  For a more nuclear
#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
# .idea/
```

I usually take the nuclear option when I'm working on something that I expect will be touched by students. ;)

cboulay:
I didn't base on dev because I noticed dev was behind main already. I'll be sure to make any future PRs against dev, but it would be great if you could rebase dev on main so that it would include the latest fixes.

Thanks for the quick turnaround!

PR #50, opened 2023-11-17
fixes #48 by popping unneeded axis out of messages.axes

### LINKED ISSUE #48: bugfix: Spectrum should pop transformed axis out of out_msg.axes
The outgoing message has `axes` field that includes the old axis (e.g. "time") even if it is no longer present because it was overwritten by "freq".

Editing [this code](https://github.com/iscoe/ezmsg/blob/dev/extensions/ezmsg-sigproc/ezmsg/sigproc/spectral.py#L126-L128) should work:

```python
        new_dims = [d for d in message.dims]
        if self.SETTINGS.out_axis is not None:
            new_dims[axis_idx] = self.SETTINGS.out_axis
            if self.SETTINGS.out_axis != axis_name:
                _ = new_axes.pop(axis_name)
```

cboulay:
See #48 for motivation.


griffinmilsap:
Good find, thanks for the contribution!!

PR #51, opened 2023-11-17
zmq - Fixes #49 by making wait for subscribers optional.

### LINKED ISSUE #49: bugfix: zmq system stalls out if there are no subscribers to ZMQSenderUnit
Sending messages to `ZMQSenderUnit` when it does not have any subscribers will cause it to go into an infinite loop. As messages accumulate, eventually the whole system will stall out.

https://github.com/iscoe/ezmsg/blob/0cbdc975daa8b8076e2c846cf6c07eefaf2407ce/extensions/ezmsg-zmq/ezmsg/zmq/units.py#L104-L107

It's fairly normal in ZMQ applications to publish when there are no subscribers. The messages simply go nowhere. I understand the use case of making sure the first few messages do not get lost while waiting for a consumer to pop up, so I get the need for some mechanism to not lose messages in some cases.

One easy short-term solution is to add a setting to `wait_for_subscriber: bool = True` and include a check for that in the `while` condition. Personally, I think default should be `False` to be consistent with normal zmq operation.

cboulay:
See #49 for motivation.
Also added support for json-encodable message-types.

### DIFF
--- a/extensions/ezmsg-zmq/ezmsg/zmq/units.py
+++ b/extensions/ezmsg-zmq/ezmsg/zmq/units.py
@@ -45,6 +45,7 @@ class ZMQSenderSettings(ez.Settings):
     write_addr: str
     zmq_topic: str
     multipart: bool = False
+    wait_for_sub: bool = True
 
 
 class ZMQSenderState(ez.State):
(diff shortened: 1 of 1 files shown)


pperanich:
I agree that we should make waiting for subscribers optional.

On the topic of json-encodeable message types, I'm not convinced we should have the default behavior json encode messages that aren't of ZMQMessage type. I would be in support of adding an optional Setting for specifying the serialization format of the incoming messages. E.g.
```python
class ZMQSenderSettings(ez.Settings):
    ...
    ser: Optional[Callable]

...

class ZMQSender(ez.Unit):
    ...
    async def zmq_subscriber(self, message: ZMQMessage):
        ...
        if self.SETTINGS.ser is not None:
            message = self.SETTINGS.ser(message)
```

So for your use case:
```python
ZMQSenderSettings(ser=lambda m: json.dumps(m).encode('utf-8'))
```

Possibly worth migrating this topic to a separate issue. I am good with the core purpose of this PR.

cboulay:
Yeah the JSON thing was just a way for me to test it and I thought it might be useful generally. But I think this kind of message-type-conversion probably belongs in its own Unit. I reverted that change.

PR #52, opened 2023-11-17
sigproc.ewm - Fixes #47 by expanding dims of scale_arr

### LINKED ISSUE #47: Enhancement: EWMA should work with arrays with ndim > 2
The following snippet right above the `def ewma` can make it work, but I had putting a while loop inside a function that gets called frequently. I'm sure there's a better way.

```python
            while scale_arr.ndim < buffer_data.ndim:
                scale_arr = scale_arr[..., None]
```

And this would have to be accompanied by removing the `, np.newaxis]` slicing on scale_arr within the function.

cboulay:
fixes #47 - scale_arr now has same ndim as buffered data, which allows for > 2-dim input.

### DIFF
--- a/extensions/ezmsg-sigproc/ezmsg/sigproc/ewmfilter.py
+++ b/extensions/ezmsg-sigproc/ezmsg/sigproc/ewmfilter.py
@@ -73,9 +73,12 @@ class EWM(ez.Unit):
             buffer_data = buffer.data
             buffer_data = np.moveaxis(buffer_data, axis_idx, 0)
 
+            while scale_arr.ndim < buffer_data.ndim:
+                scale_arr = scale_arr[..., None]
+
             def ewma(data: np.ndarray) -> np.ndarray:
-                mult = scale_arr[:, np.newaxis] * data * pw0
-                out = scale_arr[::-1, np.newaxis] * mult.cumsum(axis=0)
+                mult = scale_arr * data * pw0
+                out = scale_arr[::-1] * mult.cumsum(axis=0)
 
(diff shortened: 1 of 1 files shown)


griffinmilsap:
I'd like to have a test written before merging this one; I'll take a crack at that in the coming days.  I suspect this will function just fine, but this unit is quite complicated and used in some important places for us at the moment.

extensions/ezmsg-sigproc/ezmsg/sigproc/ewmfilter.py
```diff
@@ -73,9 +73,12 @@ async def sync_output(self) -> AsyncGenerator:
             buffer_data = buffer.data
             buffer_data = np.moveaxis(buffer_data, axis_idx, 0)
 
+            while scale_arr.ndim < buffer_data.ndim:
```
pperanich, line 76: Could also change this to
```python
if scale_arr.ndim < buffer_data.ndim:
    scale_arr = np.expand_dims(scale_arr, list(range(scale_arr.ndim, buffer_data.ndim)))
```

cboulay: Definitely happy to get rid of a while loop!
Let me know if you're interested in merging this PR in theory then I'll make the requested edits.

pperanich: It all looks good to me. I'll defer to @griffinmilsap if he'd really like to get some tests in place prior to making changes.

Issue #46, opened 2023-11-15
Request: sigproc SpectralBandpower


cboulay:
A Collection combining Window + Spectrum (or #44 Spectrogram) and #45 RangedMean. The output will be the power within each band, based on slicing and averaging the frequency axis.

I don't mind making a PR for this, pending #45, but I'd like to know if it's desired or not and if we can discuss the design a bit.

cboulay:
There are other multi-band estimators like filterbank + analytic magnitude or filterbank + square + smooth + sqrt.

It would also be great to get `specparam` in realtime but they haven't optimized that yet.

I think the FFT-based SpectralBandpower is probably the easiest to implement.

Issue #45, opened 2023-11-15
Request: RangedAgg Unit (needs better name too)


cboulay:
On the provided axis, for each range, slice the data within the range criteria and apply an operator (mean, median, sum).

I'm happy to do this as a PR but I'd like some suggestions for names and settings.

Unit name possibilities:
* RangedAgg
* RangedMean
* MultiSliceAgg
* etc.

And for the settings, when defining the ranges, should that be an ordered dictionary with keys corresponding to tick-values/labels along the new axis and values specify the range along the old axis? Or maybe we keep the axis the same and the new tick-values are the mean of the tick-values caught in the slice? The latter seems more straightforward.

griffinmilsap:
This will take a moment to digest; I haven't considered this pattern for this sort of operation yet -- Possible this "RangedAggregate" function belongs in core, or possibly even in AxisArray?  Also thinking some variety on "map" might make a better name that's consistent with similar functionality in functional programming, XArray, and Pandas.

PR #53, opened 2023-11-17
BandPower - Fixes #44, #45, and #46

### LINKED ISSUE #44: Request: sigproc Spectrogram Collection
I think it's a pretty common use-case to put a Window in front of Spectrum. This should probably be in a pre-made collection with easy-access settings.

(1-of-several requests to get at SpectralBandpower)

cboulay:
I don't expect this to be merged without major revisioin, but I hope it's a good starting point for discussing adding BandPower to sigproc.


cboulay:
Yeah I still want to work it out a bit. Using the revised Window will help. I've had many more hours with ezmsg since I opened this PR and I'm confident I can make a cleaner implementation, or at least one that's more consistent in its naming conventions and parameterization.

src/ezmsg/core/message.py
```diff
@@ -36,7 +36,13 @@ def __init__(self):
         )
 
 
-class Flag(Message):
+@dataclass
```
griffinmilsap, line 39: Good catch

griffinmilsap:
this is massive.  it looks to me like this affects synth, window, axisarray, and introduces a few new features all together.  The feature set it adds looks REALLY nice to have, but there's only so much time in the day.  I'm going to review this by merging it into dev, running some of my existing software, and seeing if anything breaks; but just as a heads up I don't have the bandwidth to review every line of this PR and expect we'd have some followup fixes as necessary.

cboulay:
Yes of course, I'm available to make updates as issues arise.

PR #65, opened 2023-11-29
Add AdaptiveStandardScaler

cboulay:
This branch uses the Generator pattern outlined in #55.

The implementation is based on River's [AdaptiveStandardScaler](https://riverml.xyz/latest/api/preprocessing/AdaptiveStandardScaler/). In fact, 2 implementations are provided and tested. The first, `scaler`, uses river directly (import is inside `scaler` function). The second, `scaler_np`, uses only numpy.

The first is about twice as fast for single-channel data but `scaler_np` is at least twice as faster for multi-channel data.

Either implementation is faster than the current EWM (at least in a 62-channel 500 Hz pipeline I was testing) and requires less memory.

### DIFF
--- a/extensions/ezmsg-sigproc/ezmsg/sigproc/scaler.py
+++ b/extensions/ezmsg-sigproc/ezmsg/sigproc/scaler.py
@@ -1,0 +1,146 @@
+from dataclasses import replace
+from typing import Generator, Callable, TypeVar
+from typing_extensions import ParamSpec
+from functools import wraps
+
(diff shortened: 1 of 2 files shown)


cboulay:
I think I would like to see `@consumer` moved out of the example pattern and directly into the core ezmsg before this is merged, then I can clean that out of the files in this PR.

griffinmilsap:
At this point, I think I'm happy to add the generator utils to core.

PR #67, opened 2023-12-05
Adds AdaptiveStandardScaler

cboulay:
The implementation is based on River's [AdaptiveStandardScaler](https://riverml.xyz/latest/api/preprocessing/AdaptiveStandardScaler/). Two implementations are provided and tested. The first, `scaler`, uses `river` directly (`import river` is hidden inside scaler function). The second, `scaler_np`, uses only `numpy`.

`scaler` is about twice as fast for single-channel data but `scaler_np` is at least twice as faster for multi-channel data.

Either implementation is faster than the current EWM (at least in a 62-channel 500 Hz pipeline I was testing) and requires less memory.

There are two tests. The first tests both functions using known input and output data from river docstrings. The second tests a system created with the AdaptiveStandardScaler unit created from gen_to_unit.


cboulay:
Shouldn't be merged as-is because it uses `gen_to_unit`. See #85 .

PR #72, opened 2023-12-21
Move to `src` layout for ezmsg and extensions, switch to `poetry`

pperanich:
The `src` layout is a modern standard that isolates our source code from other project resources. This change minimizes the risk of accidental inclusion of unwanted files in the distribution, leads to cleaner packaging and install processes, and contributes to more reliable testing practices. For our developers and contributors, this translates to a more straightforward and error-resistant development workflow.

In conjunction with the migration to a `src` layout, `ezmsg` is also excited to convert to `poetry` for our packaging and dependency management needs. Poetry is a cutting-edge tool that offers several advantages over traditional packaging systems. Read about it [here](https://python-poetry.org/)


griffinmilsap (approved):
Beautiful!

PR #74, opened 2023-12-28
gen_to_unit appends "_SIGNAL" for AxisArray streams

### LINKED ISSUE #70: Normalize on INPUT / OUTPUT or INPUT_SIGNAL / OUTPUT_SIGNAL for stream identifiers.
Most, but not all, hand-coded units use `INPUT_SIGNAL` and `OUTPUT_SIGNAL` for the main stream identifiers. However, some use vanilla `INPUT` and `OUTPUT`. Furthermore, the new `gen_to_input` uses INPUT and OUTPUT.

It would be good if all of the units that ship with ezmsg used the same stream identifiers, and if those identifiers were also the names used for `gen_to_unit`. The latter is especially important because the identifiers are not available to IDE inspection at 'develop time' / 'compile time'.

It's possible for `gen_to_unit` to know the yield and send types so a possible solution would be to use INPUT_SIGNAL / OUTPUT_SIGNAL when the type is `AxisArray` and use the vanilla INPUT / OUPUT otherwise.

cboulay:
Fixes #70 , at least partially.

### DIFF
--- a/src/ezmsg/util/gen_to_unit.py
+++ b/src/ezmsg/util/gen_to_unit.py
@@ -23,6 +23,7 @@ from typing import (
     Type,
     TypeVar,
 )
+from ezmsg.util.messages.axisarray import AxisArray
 
 if sys.version_info < (3, 12):
     from typing_extensions import dataclass_transform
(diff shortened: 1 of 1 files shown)


griffinmilsap:
The more we conform to informal standards I brainlessly adopted from LabGraph, the more I think we should document these standard somewhere...

PR #81, opened 2024-01-17
Refactor downsample as generator

cboulay:
I didn't add any new tests but all the old tests pass.


cboulay:
This Unit is simple enough that it would be a good candidate for `gen_to_unit`. However, since the Downsample Unit alread existed, using `gen_to_unit` would have been a slight regression because there would be no static type-checking available for Downsample and DownsampleSettings. That 'feature' is probably not worth the extra boilerplate code for new Units but I'll leave it in for old Units.

PR #82, opened 2024-01-17
Add AffineTransform and CommonRereference Units with tests

cboulay:
... using generator pattern


cboulay:
`weights` can be an ndarray, or a path-like to a file that works with `np.loadtxt(weights, delimiter=",")`.

We could support other delimiters or other ways of importing files into ndarray.

cboulay:
Probably shouldn't be merged as-is because it uses gen_to_unit. See #85

extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
```diff
@@ -0,0 +1,124 @@
...
+
+import numpy as np
+import ezmsg.core as ez
+from ezmsg.util.messages.axisarray import AxisArray
+from ezmsg.util.generator import consumer, GenAxisArray
+
+
+@consumer
+def affine_transform(
+    weights: Union[np.ndarray, str, Path],
```
cboulay, line 14: Path -> os.PathLike

PR #86, opened 2024-01-26
Pass Unit args and kwargs to SETTINGS if settings not provided explicitly

### LINKED ISSUE #71: REQ: If `settings=` not provided to `Unit`'s `__init__`, then pass unknown *args and **kwargs to Unit's registered Settings
To help with evangelizing, I would like to be able to say, "Writing the script is no more onerous than providing a yaml file containing a list of nodes with configurations and their connections. The script serves double-duty as your configuration file!"

The one hurdle to me saying that with confidence is the awkward pattern:

```Python
from ezmsg.my.proc import MyNode, MyNodeSettings


my_node = MyNode(MyNodeSettings(key1=value1, key2=value2))
```

I think that I can make the argument more convincingly if we didn't also need the `MyNodeSettings` class. i.e.,

```Python
from ezmsg.my.proc import MyNode


my_node = MyNode(key1=value1, key2=value2)
```

In the end, I'm hoping for something like the following:

```Python
import ezmsg.core as ez
from ezmsg.lsl.units import LSLInletUnit
from ezmsg.sigproc.butterworthfilter import ButterworthFilter
# ... other useful Units ...


def main(somepath: str = "/default/path", someswitch: str = "defualt switchvalue") -> None:
    comps = {
        "EEG_INLET": LSLInletUnit(stream_name="BrainVision RDA", stream_type="EEG"),
        "BPFILTER": ButterworthFilter(order=2, cuton=1, cutoff=70),
        # ... etc ...
    }
    ez.run(
        components=comps,
        connections=(
            (comps["EEG_INLET"].OUTPUT_SIGNAL, comps["BPFILTER"].INPUT_SIGNAL),
            (comps["BPFILTER"].OUTPUT_SIGNAL, comps["ETC"].INPUT_SIGNAL) 
        )
    )


if __name__ == "__main__":
    try:
        import typer  # instead of argparse
        typer.run(main)
    except ModuleNotFoundError:
        main()

```

If this can be achieved then it is no more complicated than yaml files I've seen in other BCI frameworks.

cboulay:
Settings are typically passed to a Unit upon instantiation as follows (e.g.)

```Python
from ezmsg.sigproc.downsample import DownsampleSettings, Downsample

ds_unit = Downsample(DownsampleSettings(axis="time", factor=10))
```

With this change, we can alternatively pass the settings parameters directly to the Unit (without the __Settings class) and they will be used to initialize the settings. i.e.,

```Python
from ezmsg.sigproc.downsample import Downsample

ds_unit = Downsample(axis="time", factor=10)
```

It's just a little quality-of-life improvement that makes the entry-point scripts cleaner.

Fixes #71

### DIFF
--- a/src/ezmsg/core/collection.py
+++ b/src/ezmsg/core/collection.py
@@ -36,8 +36,8 @@ class CollectionMeta(ComponentMeta):
 class Collection(Component, metaclass=CollectionMeta):
     """Collections can contain subunits and connect them together"""
 
-    def __init__(self, settings: typing.Optional[Settings] = None):
-        super(Collection, self).__init__(settings)
+    def __init__(self, *args, settings: typing.Optional[Settings] = None, **kwargs):
+        super(Collection, self).__init__(*args, settings=settings, **kwargs)
 
         self._components = deepcopy(self.__class__.__components__)
         for comp_name, comp in self.components.items():
--- a/src/ezmsg/core/component.py
+++ b/src/ezmsg/core/component.py
@@ -81,7 +81,7 @@ class Component(Addressable, metaclass=ComponentMeta):
     _main: Optional[Callable[..., None]]
     _threads: Dict[str, Callable]
 
-    def __init__(self, settings: Optional[Settings] = None):
+    def __init__(self, *args, settings: Optional[Settings] = None, **kwargs):
         super(Component, self).__init__()
 
         self.SETTINGS = None  # type: ignore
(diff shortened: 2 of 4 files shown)


griffinmilsap:
I believe the settings metaclass can check for the existence of a field called settings.  That's a warning at most though; I'll accept this for now, but let's make note of delivering a warning in the future.

PR #98, opened 2024-02-28
Move timestamp addtion from codec to logger/replay

pperanich:
Remove timestamp from messagecodec JSON encoder. Instead, wrap top-level objects in dict with `ts` and `obj` fields. This solves a few issues:
1. MessageEncoder assigning different timestamps for a dataclass object and any children dataclass objects
2. MessageEncoder/Decoder previously failed on messages that were not dataclasses.

### DIFF
--- a/src/ezmsg/util/messagecodec.py
+++ b/src/ezmsg/util/messagecodec.py
@@ -120,10 +120,15 @@ class MessageDecoder(json.JSONDecoder):
         return out_obj
 
 
-def message_log(fname: Path) -> typing.Generator[typing.Any, None, None]:
+def message_log(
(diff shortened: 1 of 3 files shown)


griffinmilsap (approved):
Preston and I worked on this together.  @cboulay be aware that this changes the log file format in a backward-incompatible way.  Given that we personally know affected users, we're okay with this change, but we chose to make this a PR specifically to alert you

cboulay:
I really only use the logger for temp files, and mostly only in unit tests. I don't have any old logs that I need to re-open. I'm fine with this backward incompatibility.

cboulay (approved):

PR #103, opened 2024-03-03
`windowing` generator now yields AxisArray. 

### LINKED ISSUE #100: `windowing` generator should return an AxisArray not List[AxisArray]
I'd like to change the behavaiour of the `windowing` generator method. Currently, its output is a `List[AxisArray]`. The unit then iterates over the list, yielding one at a time.

When the `newaxis` parameter is not None, segments / windows (even if only one) appear on a new axis named something like "step" or "win", and the generator return always has len=1. If `newaxis` is None then a list is returned with len for the number of windows found on this iteration.

I would like to make `newaxis` in the generator non-optional with a default value of `"win"`. This means that we always get a new "win" axis. Then, if the Unit was configured with newaxis=None, it's up to the node to multiple yields of single-window AxisArray messages without the `"win"` axis present.

The main reason for this change is that at present the `windowing` generator cannot be composed with other generator functions. A common use case that this change will simplify is a STFT where the output of `windowing` goes into `spectrum` ('time' axis becomes 'freq') and `modify_axis` to convert the 'win' axis to 'time'.

This will not change the functioning of the Window(ez.Unit) in any way, only the generator method.

Pleaset let me know what you think.

cboulay:
Fixes #100 

The `windowing` generator now _always_ yields an `AxisArray`.

Previously, the return type was `List[AxisArray]`. Depending on the configuration, the list might have always been length 1 (in 1:1 mode) or usually 1, unless a big chunk came in and multiple windows were available.
Now, there is always a `newaxis` added ("win" by default) and that will usually have length 1, but sometimes more.

The "win" axis might also have length 0 when the most recent `send` was inadequate to fill the next window. Previously, an unfilled `.send` would yield `None` (previous yield type incorrectly failed to include `Optional[...]`). This is probably the biggest change here and might merit some discussion:
_What should generators yield when the `.send` does not provide enough data to yield a result?_ `None`, an empty `AxisArray`, or an `AxisArray` with most details correct except one dimension is `0` and `msg.data.size==0`?

The Unit still has the option for the setting `newaxis=None`, in which case the value yielded by the generator is iterated over the "win" axis. In cases where the "win" axis is length 0, nothing is iterated (i.e., nothing published).

I had to make a change to the generator-level tests to drop yielded msgs with size 0, whereas previously I was dropping yielded msgs that were None.

The system-level tests are unchanged and continue to pass. If you have a use case that doesn't work with this PR then please create a test for that and I will try to figure out a solution.

### DIFF
--- a/extensions/ezmsg-sigproc/src/ezmsg/sigproc/sampler.py
+++ b/extensions/ezmsg-sigproc/src/ezmsg/sigproc/sampler.py
@@ -97,8 +97,8 @@ def sampler(
                 continue
 
             # Check that period is compatible with buffer duration.
-            max_buf_len = int(buffer_dur * fs)
-            req_buf_len = int((_period[1] - _period[0]) * fs)
+            max_buf_len = int(np.round(buffer_dur * fs))
+            req_buf_len = int(np.round((_period[1] - _period[0]) * fs))
             if req_buf_len >= max_buf_len:
                 ez.logger.warning(
                     f"Sampling failed: {period=} >= {buffer_dur=}"
@@ -164,7 +164,7 @@ def sampler(
                 t_start = trig.timestamp + trig.period[0]
                 if t_start >= buffer_offset[0]:
                     start = np.searchsorted(buffer_offset, t_start)
-                    stop = start + int(fs * (trig.period[1] - trig.period[0]))
+                    stop = start + int(np.round(fs * (trig.period[1] - trig.period[0])))
                     if buffer.shape[axis_idx] > stop:
                         # Trigger period fully enclosed in buffer.
                         msg_out.append(
--- a/extensions/ezmsg-sigproc/src/ezmsg/sigproc/spectrogram.py
+++ b/extensions/ezmsg-sigproc/src/ezmsg/sigproc/spectrogram.py
@@ -4,7 +4,7 @@ import numpy as np
 
 import ezmsg.core as ez
 from ezmsg.util.messages.axisarray import AxisArray
-from ezmsg.util.generator import consumer, GenAxisArray  # , compose
+from ezmsg.util.generator import consumer, GenAxisArray, compose
 from ezmsg.util.messages.modify import modify_axis
 from ezmsg.sigproc.window import windowing
 from ezmsg.sigproc.spectrum import (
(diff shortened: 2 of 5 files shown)


cboulay:
Oops. I did't meant to base on `main`. I'll let the tests complete then I'll rebase on `dev`.

cboulay:
@griffinmilsap , last time I changed windowing it broke one of your pipelines. Can you please check this one?

griffinmilsap:
`Window` is such a beast.  Easily one of our most complicated and heavily used additions to the signal processing library.  I can say that depending on where this `win` dimension is placed, this will probably break some of my existing work, especially where (by happenstance) some of my pipelines actually already rely on a `win` dimension.  How would you feel about changing `win` to something like `_new_axis`?

The idea of 0-length dimensions in `AxisArray` is also something I never really considered.  It looks like the `Unit`ization of this generator drops 0-length messages so this change won't result in higher messaging overhead for no reason, so that's good.  In general, I'm in favor of retiring `Optional` data types because it makes linting easier and we don't have to check for `is None` later.

 @pperanich do we rely on `ezmsg.sigproc.window`, or its generator implementation in any of the HLM or OI work?  If not, given that this change only effects the generator, I'm for merging this if we can address the issue above.  

As always, @cboulay thank you for the contributions; these are incredibly valuable and I apologize it took so long to get to this one.

cboulay:
(rebased)

@griffinmilsap , I chose "win" because I thought that's what you were using but I thought it was scoped to exactly using a windowing node. I didn't think about what would happen if you applied the Window unit to an AxisArray that already had a "win" axis.

That is, if you tell the Window _Unit_ that you don't want a newaxis and the generator inserts the "win" axis anyway for its own internal bookkeeping, it will be confused by the existing "win" axis.

Sorry about that.
I'm happy to rename the default bookkeeping axis to something else. How is `"_windowing"`? Ultimately it only matters for direct users of the generator, as in some offline pipelines.

Reminder to myself: Modify my offline pipelines to be explicit about the `newaxis="win"` argument.

cboulay:
I accidentally had an unrelated commit sneak into this PR. It's similar in that it uses size-0 arrays in offline pipelines, but it's for a different processing step. I'll try to separate them.

griffinmilsap:
I think its time to merge this and if I run into issues on dev, I've got time to react to them :)

PR #117, opened 2024-04-21
Give CAR passthrough mode.

cboulay:


griffinmilsap:
Seems benign enough :)

Issue #43, opened 2023-10-26
How to track channel labels and locations?


cboulay:
If I have a source Unit that has information about channel labels and locations, what is the recommended way to relay that to downstream Units? It's too much information to put into a e.g. `SpaceAxis` and transmit with every message.

So do I add a META_OUTPUT port to my Unit and transmit it once when I get that information and again whenever it changes? Is there any way to guarantee that this port is transmitted first so that downstream Units will receive the updated metadata before receiving streaming data with the new format?

griffinmilsap:
In practice, I subclass AxisArray and add custom fields to it that track the extra information.  I tend to take it on the nose serializing that info with every message with the understanding that ezmsg is ... _very_ .. fast.  Seems inefficient, but it is the vastly simpler solution.  Given ezmsg's message cacheing under the hood, the impact is actually quite minimal.

If you really want to do this with a meta_output stream, I recommend setting up receivers for your meta_output that wait until that metadata is received before processing any EEG messages.  I can post an example impl if you're interested

griffinmilsap:
As a matter of fact, if you do it like this:
``` python
@dataclass
class EEGMessage(AxisArray):
    ch_names: typing.List[str]
    ch_locs: npt.ndarray
```

Under the hood, message passing is implemented with only one copy (with zero-copy reads) for any data type that uses the array API.  It might just be faster than you think ;)

griffinmilsap:
closing for now, but this issue should probably be referenced in a future "Patterns" or FAQ page (#54)

PR #123, opened 2024-05-11
Support a "labels" field in AxisInfo if present. 

cboulay:
ezmsg-lsl has a [SpaceAxis that I patch into AxisArray](https://github.com/labstreaminglayer/ezmsg-lsl/blob/62d1458bb335b82d7d7b5c7930d5770c4463ee70/src/ezmsg/lsl/units.py#L13-L23), and SpaceAxis has a `labels` field. This solves an issues that we discussed in #43 

Units that transform the AxisArray in a way that modifies the length of the dimension with the "labels" field, should probably also modify the labels, if present! This PR does that for Units that I'm aware of that modify the shape of the array along axes other than "time".

For `affine_transform`, it's difficult to know what the new labels should be so in most cases they are just dropped. However, if the `affine_transform` has rows or columns of zeros then we can get a good guess at thew new labels.

### DIFF
--- a/extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
+++ b/extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
@@ -39,6 +39,9 @@ def affine_transform(
         weights = weights.T
     weights = np.ascontiguousarray(weights)
 
+    b_first = True
+    new_xform_ax: typing.Optional[AxisArray.Axis] = None
+
     while True:
         axis_arr_in = yield axis_arr_out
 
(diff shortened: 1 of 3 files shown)


griffinmilsap:
I didn't know about an lsl connector; this is really cool!  When you're ready, let's add it into the README as another extension :D

AxisArray is VERY inspired by [XArray](https://github.com/pydata/xarray) (my absolutely favorite data structure for offline analyses) that unfortunately has a little too much going on under the hood to be performant enough for low latency, high-message rate stream processing.  An earlier version of the ezmsg-sigproc lib actually used DataArrays for the messages and that was a lovely developer experience even if my CPUs pegged processing 8 channels of 250 Hz EEG data -- but I digress.

I feel like the general solution to this particular problem is adding a `coords` implementation into AxisArray that captures some core set of functionality that we find in XArray's DataArray `coords`.  I wrote the majority of the AxisArray implementation in an afternoon when I was pressured by an incoming deadline.  I spent about an hour thinking about how to get `coords` in without destroying performance or my guiding mantra that "AxisArray is simply a metadata wrapper around a numpy array".  I came to the conclusion at that time that a limited `coords` functionality in AxisArray would 1. be incredibly useful, 2. could be performant if we removed all the integrity checks/safety rails that XArray's implementation has.  I simply didn't have the bandwidth (at that time) to even do a first-pass implementation.

In summary, AxisArray needs some TLC for 
1. Better Typehinting (even simple fixes like sel/isel returning subclasses of AxisArray, not just AxisArrays)
2. Better dev experience/consistency (why is concatenate a `staticmethod`?) 
3. A lightweight `coords` implementation

__Coming back to today and this PR:__ These operations on coords/labels probably eventually belong in AxisArray helper functions/implementation rather than spread throughout the signal processing modules.  Collecting common operations on AxisArrays together in one place would make this whole thing a lot easier to maintain, but if this gets you where you're going _today_ I support a merge after review (feel free to merge it in yourself once my review posts)

extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
```diff
@@ -48,6 +51,43 @@ def affine_transform(
...
+                    elif np.all(b_used_inputs):
+                        # Strange scenario: New outputs are filled with empty data.
+                        in_ix = 0
+                        new_labels = []
+                        for out_ix in range(n_out):
+                            if b_filled_outputs[out_ix]:
+                                new_labels.append(in_labels[in_ix])
+                                in_ix += 1
+                            else:
+                                new_labels.append("")
```
griffinmilsap, line 85: Maybe append None here instead?  We could also assume `labels` will always be a `typing.List[str]` if it exists

griffinmilsap, line 85: Maybe append `None` here instead? If you keep `""`, we should be explicit that we assume labels will always be a `typing.List[str]` if it exists

griffinmilsap (commented):
Looks alright to me :). Maybe consider adding your monkeypatch into AxisArray while you're at it.

griffinmilsap (approved):

extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
cboulay: Yes, it is always a `typing.List[str]`, not a `typing.List[typing.Optional[str]]`. The `""` is intended to indicate that we don't know what the label is. If anything, I was thinking about replacing the `""` with `"?"` or even `str(len(new_labels)+1)`. But I decided against these because integers might already have a different meaning, and I just know that "?" is going to interfere with somebody's localization or other parser down the line eventually.

cboulay:
Moved to https://github.com/ezmsg-org/ezmsg-sigproc/pull/1

PR #126, opened 2024-06-09
Fix downsample

cboulay:
* New unit tests for core (generator function) processing
* accommodate size-0 arrays
* when there is no output, return size-0 arrays instead of None
* slight optim. using template struct -- this approach is abandoned in a future PR so don't dwell on it.

ETA:
This also fixes a bug! Prior to this PR, the downsample generator would not make proper copies and would overwrite shared fields in long-passed messages.


cboulay:
Because this fixes a bug that I created, I'm inclined to go ahead and merge this myself.
However, the bug is fairly low priority because it only affects people who were using downsample generator, outside the ez.run() context (which did deepcopy on each msg).

Issue #124, opened 2024-06-03
Efficient and safe mutation of AxisArray


cboulay:
I occasionally run into issues when manipulating AxisArray objects when I fail to properly break the link between the incoming message and the newly generated output. A message will get manipulated in unexpected ways because I end up modifying a shared dictionary or other mutable object. This happens exclusively when using the (generator) transformation methods without the ezmsg runner, because the runner does a deepcopy of every message unless told not to.

So I find myself re-inventing the solution to this problem over and over and I've gotten it wrong a couple times. For example, `downsample` has a bug that I've fixed in cboulay/working but I haven't pade a PR for yet. Furthermore, it turns out that my most common minimal-code-lines solution is slower than the more straightforward solution.

I wrote a notebook for myself that goes through the issue in detail so hopefully I'll never make the same mistake again, especially if I refer back to this notebook. I'll add it into the repo somewhere eventually. For now it is attached here.
[optim_axis_array_mods.ipynb.zip](https://github.com/user-attachments/files/15528336/optim_axis_array_mods.ipynb.zip)

Something surprising came out of this notebook: `replace` is about twice as slow as creating a new object. I wasn't expecting that.

```Python
def full_modify_1(in_msg):
    msg_cls = type(in_msg)
    out_msg = msg_cls(**in_msg.__dict__)
    ax_cls = type(in_msg.axes["time"])
    out_msg.axes = {
        k: (v if k != "time" else ax_cls(gain=in_msg.axes["time"].gain, offset=-88.88))
        for k, v in in_msg.axes.items() if k != "time"
    }
    return out_msg


def full_modify_2(in_msg):
    return replace(
            in_msg,
            axes={
                **in_msg.axes,
                "time": replace(
                    in_msg.axes["time"],
                    offset=-99.99
                )
            }
        )
    

%timeit [full_modify_1(_) for _ in in_msgs]
%timeit [full_modify_2(_) for _ in in_msgs]
```

```
1.91 ms ± 6.55 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
4.51 ms ± 16.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Or 0.9 vs 2.2 us per message (2048 messages per loop). This doesn't sound like much, but if you have 20-50 nodes that manipulate the axes then this will compound.

I'll make an optimization pass when I'm back from the conference, then I'll set `zero_copy=True` on all the nodes that I can.

[...]

PR #127, opened 2024-06-09
sigproc optimization

cboulay:
Note that the first 2 commits in this PR are already in other PRs in #123 and #127. I included them here because there would be conflicts otherwise. If those 2 PRs get merged then this PR will be a little easier to parse.

What the 3rd commit does:

Following my comments in #124 , I went through all of the sigproc modules and refactored them to create new msg objects because it turns out that shallow-copy + modification is faster than `replace`. This is shown explicitly in the new notebook added to `extensions/tests/resources/`.

Then, confident that we're now making proper copies and not mutating shared objects, I turned on `zero_copy=True` for as many of these Units as I could.

### DIFF
--- a/extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
+++ b/extensions/ezmsg-sigproc/src/ezmsg/sigproc/affinetransform.py
@@ -1,3 +1,4 @@
+import copy
 from dataclasses import replace
 import os
 from pathlib import Path
@@ -92,7 +93,7 @@ def affine_transform(
 
         if data.shape[axis_idx] == (weights.shape[0] - 1):
             # The weights are stacked A|B where A is the transform and B is a single row
-            #  in the equation y = Ax + B. This supports NeuroKey's weights matrices.
+            #  in the equation y = Ax + B. We append a column of ones to our data.
             sample_shape = data.shape[:axis_idx] + (1,) + data.shape[axis_idx+1:]
             data = np.concatenate((data, np.ones(sample_shape).astype(data.dtype)), axis=axis_idx)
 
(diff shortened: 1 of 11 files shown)


cboulay:
Maybe worth mentioning, I decided to use `new_msg = copy.copy(old_msg)` instead of `new_cls = type(old_msg); new_msg = new_cls(old_msg.__dict__)` despite the former being a bit slower, because it's more readable and somehow I think the use of `__dict__` is more fragile though I don't have anything to back that up.

cboulay:
I just saw some weird behaviour when trying to use this approach to optimize the BandPower node (which isn't included in this PR, but its dependencies are). I'm modifying this to a draft while I add some more unit tests and track down the source of the problem.

cboulay:
The weird behaviour was that I attempted to override a parent method when doing `@ez.subscriber(OUTPUT_SIGNAL, zero_copy=True)` but I misspelled the function name and ended up double-processing the data. I've since solved this in my cboulay/working branch. I'll rebase and redo the commit here in the near future.

griffinmilsap:
Another thought; shallow copy with modification of the dataclass may be faster than replace, but replace calls the constructor, and importantly, the `__post_init__` method of a dataclass which is important for some of my messages.  If the speedup is that substantial, we should consider whether its worth losing that expected (?) behavior

cboulay:
Yeah the gains aren't that big in practice, at least not on the scale I'm working on so far. I am looking forward to 100x more throughput than I'm doing currently though but maybe I'll need custom CUDA nodes for that.

I'll close this PR and open new ones just for the zero_copy=True changes and some unit tests to check that the input msg has not been modified relative to a deepcopy backup.

BTW, do you need to use `replace` on the axis objects too, or just the message itself?
i.e., Is the following OK?

```Python
new_axes = {**old_axes, "time": TimeAxis(fs=fs, offset=new_offset)}
replace(in_msg, data=new_data, axes=new_axes)
```

Or do you need

```Python
new_axes = {**old_axes, replace(old_axes["time"], offset=new_offset)}
replace(in_msg, data=new_data, axes=new_axes)
```

?

griffinmilsap:
I'm not doing anything weird with the axes at this point, but I could see implementing custom axes in the future.  It might be worthwhile to do the latter over the former.

Issue #134, opened 2024-07-15
Inherited Unit does not inherit State correctly on python versions > 3.9


pperanich:
I recently ran into an odd bug where an inherited unit does *not* inherit the base classes `STATE` but does inherit it's `SETTINGS`. Importantly, this bug is only present in Python versions greater than 3.9. See below for a minimal example.

```python
import ezmsg.core as ez


class BaseSettings(ez.Settings):
    y: int = 10


class BaseState(ez.State):
    x: int = 9


class BaseUnit(ez.Unit):
    SETTINGS: BaseSettings
    STATE: BaseState

    async def initialize(self) -> None:
        ez.logger.info(f"{self.SETTINGS=}, {self.STATE=}")


class ChildUnit(BaseUnit):
    async def initialize(self) -> None:
        ez.logger.info("In child")
        ez.logger.info(f"{self.SETTINGS=}, {self.STATE=}")
        return await super().initialize()


if __name__ == "__main__":
    ez.run(ChildUnit())
```

Output on 3.9:
```
2024-07-15 13:18:26.014 - pid: 35744 - TaskThread - INFO - initialize: In child
2024-07-15 13:18:26.014 - pid: 35744 - TaskThread - INFO - initialize: self.SETTINGS=BaseSettings(y=10), self.STATE=BaseState(x=9)
2024-07-15 13:18:26.014 - pid: 35744 - TaskThread - INFO - initialize: self.SETTINGS=BaseSettings(y=10), self.STATE=BaseState(x=9)
2024-07-15 13:18:26.016 - pid: 35744 - MainThread - INFO - run: All processes exited normally
```

Output on 3.10:
```
2024-07-15 13:18:25.996 - pid: 35743 - TaskThread - INFO - initialize: In child
2024-07-15 13:18:25.996 - pid: 35743 - TaskThread - INFO - initialize: self.SETTINGS=BaseSettings(y=10), self.STATE=State()
2024-07-15 13:18:25.996 - pid: 35743 - TaskThread - INFO - initialize: self.SETTINGS=BaseSettings(y=10), self.STATE=State()
2024-07-15 13:18:25.997 - pid: 35743 - MainThread - INFO - run: All processes exited normally
```

pperanich:
Fixed by #135

PR #135, opened 2024-07-18
fix: Settings and State assigned instead of annotated.

pperanich:
This PR was orginally intended to simply fix https://github.com/ezmsg-org/ezmsg/issues/134, but I believe this is a good time to refactor the specification of `SETTINGS` and `STATE` in a `Component` to be more pythonic. Thus, this PR enables and encourages the declaration of `SETTINGS` and `STATE` via assignment rather than by annotation. The annotation approach will not be deprecated immediately, but will emit a logging warning to notify the end-user of this design change.

This fixes static type-check errors introduced by annotating.

You may run the following command to fix downstream codebases using the `find` and `sed` tools:

```bash
find . -type f -name "*.py" -exec sed -i 's/SETTINGS:/SETTINGS =/g' {} +
find . -type f -name "*.py" -exec sed -i 's/STATE:/STATE =/g' {} +
```

### DIFF
--- a/src/ezmsg/core/component.py
+++ b/src/ezmsg/core/component.py
@@ -25,13 +25,13 @@ class ComponentMeta(ABCMeta):
         if "SETTINGS" in cls.__annotations__:
             settings_cls = cls.__annotations__["SETTINGS"].__name__
             logger.warning(
-                f"{cls.__name__} SETTINGS should be assigned rather than annotated: `SETTINGS = {settings_cls}` -> `SETTINGS = {settings_cls}`"
+                f"{name} SETTINGS should be assigned rather than annotated: `SETTINGS = {settings_cls}` -> `SETTINGS = {settings_cls}`"
             )
             fields["SETTINGS"] = cls.__annotations__["SETTINGS"]
         if "STATE" in cls.__annotations__:
             state_cls = cls.__annotations__["STATE"].__name__
             logger.warning(
-                f"{cls.__name__} STATE should be assigned rather than annotated: `STATE = {state_cls}` -> `STATE = {state_cls}`"
+                f"{name} STATE should be assigned rather than annotated: `STATE = {state_cls}` -> `STATE = {state_cls}`"
             )
             fields["STATE"] = cls.__annotations__["STATE"]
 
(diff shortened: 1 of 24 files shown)


src/ezmsg/core/component.py
```diff
@@ -9,76 +9,98 @@
...
-            ):
-                cls.__state_type__ = getattr(base, "__state_type__")
+        cls.__streams__ = {}
 
         if "SETTINGS" in cls.__annotations__:
-            cls.__settings_type__ = cls.__annotations__["SETTINGS"]
-
+            settings_cls = cls.__annotations__["SETTINGS"].__name__
+            logger.warning(
+                f"{cls.__name__} SETTINGS should be assigned rather than annotated: `SETTINGS = {settings_cls}` -> `SETTINGS = {settings_cls}`"
```
cboulay, line 28: f"{cls.__name__} SETTINGS should be assigned rather than annotated: `SETTINGS: {settings_cls}` -> `SETTINGS = {settings_cls}`"

cboulay:
I'm a bit confused by this. Am I correct in my understanding that `self.SETTINGS` is an object but `cls.SETTINGS` is a type?

pperanich:
Yes, you are correct. I am open to alternative approaches to specifying `SETTINGS` and `STATE` as well, but I think this change is still a step in the right direction.

The crux of the issue is that with the annotation approach, I get the following diagnostic from my IDE when specifying `SETTINGS` and `STATE` for every Unit.
```
1. "SETTINGS" overrides symbol of same name in class "Component"
     Variable is mutable so its type is invariant
       Override type "CountSettings" is not the same as base type "Settings" [reportIncompatibleVariableOverride]
```

src/ezmsg/core/component.py
pperanich: Updated to use `name` instead.

cboulay: Sorry, that was just GitHub messing up the formatting of my comment. I thnk `__name__` part is fine. I was commenting on the first part (to be changed) reading `SETTINGS =` while I think it should be `SETTINGS:` that you want to change.

pperanich: Ahh, I totally missed that, thanks.

griffinmilsap:
@cboulay Preston and I have chatted about this quite a bit, and beyond the linter issues, I keep coming back to how weird it is that the user instantiates sub-units, which get deep-copied per class instance, but the settings and state are simply type-hinted; and the code reads the annotations to create bespoke settings and state objects.  I think the reason LabGraph worked this way was related to their messaging backend (cthulhu) but I think consistency is more important here.  There's no shortage of 'magic' in ezmsg, but using typehints to drive functionality of the code is one of the biggest bits of magic that is arguably a big no-no; as annotations/typehints are pretty exclusively supposed to only provide typing data to linters and make code more readable.

Preston went through this changeset with me and although I'll have to patch a bunch of code, provided this is all functional and tests pass - I think it's a good change

cboulay:
I agree it's a step in the right direction. This is something that tripped me up once for 30 minutes -- I accidentally used assignment instead of annotation, or was it vice versa?

I'm not suggesting we do the following, but I'm curious ... would it be better or worse to have something like `cls.SETTINGS_TYPE` or `cls.TSettings` and then use magic to init `obj.settings` (lower case)? Definitely a lot more editing required, many places, but it would eliminate this particular ambiguity.

cboulay:
I'm ready for this to be merged in if you are. Please go ahead.

pperanich:
> I agree it's a step in the right direction. This is something that tripped me up once for 30 minutes -- I accidentally used assignment instead of annotation, or was it vice versa?
> 
> I'm not suggesting we do the following, but I'm curious ... would it be better or worse to have something like `cls.SETTINGS_TYPE` or `cls.TSettings` and then use magic to init `obj.settings` (lower case)? Definitely a lot more editing required, many places, but it would eliminate this particular ambiguity.

The biggest issue with this approach is that we would no longer get any type hinting for `self.SETTINGS` or `self.STATE`. I'll move forward with what I had originally proposed for now.

PR #136, opened 2024-07-24
fix: Linting fixes and cleanup

pperanich:
This PR is intended just to cleanup the repo and fix a variety of linting errors, checked via `ruff`

### DIFF
--- a/src/ezmsg/core/backendprocess.py
+++ b/src/ezmsg/core/backendprocess.py
@@ -16,7 +16,7 @@ from contextlib import suppress, contextmanager
 from concurrent.futures import TimeoutError
 
 from .stream import Stream, InputStream, OutputStream
-from .unit import Unit, TIMEIT_ATTR, PUBLISHES_ATTR, SUBSCRIBES_ATTR, ZERO_COPY_ATTR
+from .unit import Unit, TIMEIT_ATTR, SUBSCRIBES_ATTR, ZERO_COPY_ATTR
 
 from .graphcontext import GraphContext
 from .graphserver import GraphService
@@ -24,6 +24,7 @@ from .shmserver import SHMService
 from .pubclient import Publisher
 from .subclient import Subscriber
 from .messagecache import MessageCache
+from abc import abstractmethod
 
 from typing import (
     List,
@@ -31,6 +32,7 @@ from typing import (
     Callable,
     Any,
     Optional,
+    Sequence,
     Set,
     Coroutine,
     DefaultDict,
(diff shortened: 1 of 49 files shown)


src/ezmsg/util/perf_test.py
```diff
@@ -91,8 +93,8 @@ class LoadTestSubscriberState(ez.State):
 
 class LoadTestSubscriber(ez.Unit):
     INPUT = ez.InputStream(LoadTestSample)
-    SETTINGS: LoadTestSettings
-    STATE: LoadTestSubscriberState
+    SETTINGS =LoadTestSettings
```
cboulay, line 96: Should `SETTINGS` and `STATE` really be "=" here? I thought they were supposed to be `:`?

tests/messages/test_axisarray.py
```diff
@@ -121,19 +126,22 @@ def test_sel():
         data, dims=["dim0"], axes=dict(dim0=AxisArray.Axis(gain=gain, offset=offset))
     )
 
-    aa_sl = aa.sel(dim0=slice(None, -10.75, 1.5))  # slice based on axis info
-    aa_idx = aa.isel(dim0=-1) # index slice of last index
+    aa.sel(dim0=slice(None, -10.75, 1.5))  # slice based on axis info
```
cboulay, line 129: `aa.sel(...)` returns something. The better fix here would be to assert something on the return rather than to discard it.

cboulay (commented):
I added a few inline comments to address.

src/ezmsg/util/perf_test.py
cboulay: Ah, I see this is the topic of #135

pperanich: Apologies, this was not supposed to creep into this PR.

src/ezmsg/core/backendprocess.py
```diff
@@ -172,19 +177,22 @@ async def setup_state():
...
         try:
             logger.debug("Waiting at start barrier!")
             self.start_barrier.wait()
 
-            threads = [
-                loop.run_in_executor(None, thread_fn, unit)
-                for unit in self.units
-                for thread_fn in unit.threads.values()
-            ]
+            # FIXME: Threads not currently supported
```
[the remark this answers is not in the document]
pperanich: Removed this patch from the commit.

tests/messages/test_axisarray.py
pperanich: Added an assert to check contents of `aa.sel(...)`.

cboulay:
LGTM. Please merge if ready.

pperanich:
I've added a pre-commit-config to always run the linter and formatter prior to commits. @griffinmilsap let me know if you are good with the format changes, or would prefer to keep this PR limited to just linting fixes.

PR #137, opened 2024-08-21
ModifyAxis can drop len==1 dims

cboulay:
ModifyAxis can drop len==1 dims if the name of the dim is provided to the map as {old_name: None}.


cboulay:
This PR augments only a single Unit that I added not too long ago, and it doesn't break the previous API, so I think there probably isn't really any concern about this breaking anything. If there are no objections then I'll merge this in tomorrow.

PR #139, opened 2024-08-22
Switch package management from poetry to uv

pperanich:
I've been using `rye` as my go-to tool for managing Python toolchains, projects, and virtual environments for some time. `uv` was initially used under the hood in `rye`, but it now also supports direct project management. Overall, `uv` offers a significant speed advantage over `poetry` for dependency resolution and installation, with the added bonus of managing Python versions as well. While this isn't strictly necessary, it would be a nice enhancement.

Another benefit is that the pyproject.toml does not need special configuration sections as it does in poetry. Instead, it relies on the standards detailed in [the Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/), with PEP508 dependency specifications.

For more details, you can refer to the [uv documentation](https://docs.astral.sh/uv/).


cboulay:
I only advocated for poetry before because it seemed to be the _de facto_ environment manager for package authors, even though I kind of didn't like it. If you like this one better then I'm all for switching.

The only small regression is that I use PyCharm and it can auto-detect poetry environments... but it's not great at it so it's really not a huge loss.

I approve this PR.

cboulay:
I tried to add ezmsg-rerun to an existing project using `poetry add path/to/ezmsg-rerun` (whether local or git) and kept facing `Unable to create package with no name`.

I decided to try the same with this branch and I couldn't add it. I hope switching to uv doesn't meant that poetry users can no longer add ezmsg.

cboulay:
Needed `poetry self add pkginfo==1.10.0` then it's adding again.

pperanich:
From [here](https://github.com/python-poetry/poetry/issues/9191) this looks to be potentially a poetry bug?

cboulay:
Yeah. It looks like it was fixed in poetry 1.8.
It's also a pretty easy problem to solve with a tiny bit of googling so I don't think we need to do anything proactive about it. Please carry on ;)

PR #143, opened 2024-08-31
Update sigproc docs

cboulay:
Adds math module and submodules.

Adds blurb about ezmsg.sigproc mostly using AxisArray messages and those sometimes having .data of size==0. See https://github.com/ezmsg-org/ezmsg-sigproc/issues/5


cboulay:
This is just an update to docs that I wrote so I don't think anyone will object. Merging.

PR #144, opened 2024-09-02
Add `array_chunker` to iterate over chunks of arbitrary npt.ArrayLike objects

cboulay:
This is useful for tests, examples, and quick offline analyses with small'ish datasets.


cboulay:
Alternatively, we could create a large AxisArray then add a new `chunk_over_axis` method to `AxisArray`.

PR #145, opened 2024-09-02
Add `key` field to AxisArray structure, with a Unit to set the key and another to filter on key

### LINKED ISSUE #142: [REQ] key field on data messages (AxisArray or other)
I'm increasingly in the need of a `key` field on my AxisArray messages.

This field could be populated by source nodes (e.g., {topic-name} for ZMQ sub, or {stream_name}-{stream_type} for LSLInlet), or it could be set / overridden by a `SetKey` node.

Then we could have a gating node like `FilterOnKey` that only allows messages through that match the pattern.

For my present use case, this will allow me to yield messages from multiple data sources to a single OUTPUT_SIGNAL stream while iterating over a single multimodal file (e.g., XDF or NWB). Also, it will allow my multi-modal file-writers (currently NWBSink) to receive multiple data streams into it INPUT_SIGNAL and write those all to the same file.

I've had other needs for this feature in the past that I can't recall now.

cboulay:
(ignore array_chunker commits)

Closes #142 

This PR adds a `.key` field to AxisArray then adds a couple functionalities related to it:

* `AxisArray.concat(..., filter_key="some key")` will only concatenate messages with matching keys
* `AxisArray.concat(..., filter_key=None)` (default) will not filter but will set key to `""` (default) if keys are not matching.
* `set_key` generator or `SetKey` Unit sets the .key field to a new value.
* `FilterOnKey` (Unit only) only allows through messages with matching .key value.

### DIFF
--- a/src/ezmsg/util/messages/axisarray.py
+++ b/src/ezmsg/util/messages/axisarray.py
@@ -23,6 +23,7 @@ class AxisArray:
     data: npt.NDArray
     dims: typing.List[str]
     axes: typing.Dict[str, "AxisArray.Axis"] = field(default_factory=dict)
+    key: str = ""
 
     T = typing.TypeVar("T", bound="AxisArray")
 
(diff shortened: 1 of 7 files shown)


cboulay:
Close in favour of #146

PR #146, opened 2024-09-04
Add key field to AxisArray structure, with a Unit to set the key and another to filter on key

cboulay:
This PR adds a `.key` field to AxisArray then adds a couple functionalities related to it:

* `AxisArray.concat(..., filter_key="some key")` will only concatenate messages with matching keys
* `AxisArray.concat(..., filter_key=None)` (default) will not filter but will set key to `""` (default) if keys are not matching.
* `set_key` generator or `SetKey` Unit sets the .key field to a new value.
* `FilterOnKey` (Unit only) only allows through messages with matching .key value.

This is a redo of #145 without the ArrayChunker stuff.


cboulay:
I will merge this because it was agreed on ahead of time. I'm willing to make any requested changes that might come after the merge.

PR #153, opened 2024-11-08
Feature: AxisArray with CoordinateAxes

griffinmilsap:
Added a "CoordinateAxis" capable of containing "coordinates" along axes, supporting functionality like discrete channel labels.

I realized that `axes` and my proposed `coords` were already quite compatible with eachother.  An example AxisArray with coordinates might look like this:

``` python
import numpy as np
from ezmsg.util.messages.axisarray import AxisArray

aa = AxisArray(
    data = np.arange(8).reshape((4,2)),
    dims = ['time', 'ch'],
    axes = {
        'time': AxisArray.TimeAxis(fs = 100.0),
        'ch': AxisArray.CoordinateAxis(
            data = np.array(['ch1', 'ch2']), dims = ['ch']),
        'interpolated': AxisArray.CoordinateAxis(
            data = np.array([True, False, True, False]), dims = ['time']),
    }
)
```

Axes that share a name with a dimension are assumed to be single-dimensional coordinate or linear axes that describe a dimension.  Axes that don't share a name with a dimension are "coordinates" and must be specified as a CoordinateAxis with a named dimension that corresponds to one of the `dims` in the AxisArray.  Currently, this syntax doesn't permit specifying a LinearAxis as a coordinate axis because there's no way to add the dim name that it corresponds to.  I can't currently think of a purpose for a non-coordinate axis as an additional named coordinate, and even if I could, I'm unsure how we would support multi-dimensional LinearAxes in a single coordinate.

This ended up looking a LOT different than I had originally conceived, but I think I'm happier with where it landed.  I'd like some more review on this before merging to dev -- but I THINK everything here is backwards compatible (some linting/typehinting might be different, but everything should be functional)


griffinmilsap:
Additionally -- this is the subset of additions that I believe are backward-compatible, but I'd like to consider retiring `AxisInfo` (I think its silly) and actually taking a real spin on `sel`ections with proper typehinting in a future PR.

cboulay:
I had to rename `np.float_` to `np.float64` because the former has been removed from newer versions of numpy. With that one change, all of the unit tests in ezmsg-sigproc are passing. I pushed that commit here. Let me know if you need me to revert.

I'm going to do some usability testing of CoordinateAxis within ezmsg-sigproc before providing more feedback.

cboulay:
I used to be able to do `msg1.axes[k] == msg2.axes[k]`. That no longer works. From the dataclass `__eq__` docs:

> This method compares the class as if it were a tuple of its fields, in order.

Either I can modify my unit tests to explicitly check `(msg1.axes[k].data == msg2.axes[k].data).all()` or we can override `__eq__` for these classes.

I'm trying that approach now locally.

```Python
    def __eq__(self, other):
        if not isinstance(other, CoordinateAxis):
            return False
        return all(
            v == getattr(other, k) if k != "data" else (v == getattr(other, k)).all()
            for k, v in self.__dict__.items()
        )
```

cboulay:
I went through all of ezmsg.sigproc to do a first pass at using this functionality. The [diff is here](https://github.com/ezmsg-org/ezmsg-sigproc/compare/dev...coords).

Other than the above comments, my only concern is that there are quite a few sigproc methods that need to know the sample rate of the incoming message, so they will only work if the selected `axis in msg.axes` (that req already existed) AND the `msg.axes[axis]` entry is of type `AxisArray.LinearAxis` (new req). My solution is to add a note to the docstring and live with the fact that some of these nodes will fail if someone tries to operate on a CoordinateAxis.

`windowing` and `sampler` currently require LinearAxis though that can be worked around now that we have CoordinateAxis to monitor. I'm tracking that issue [here](https://github.com/ezmsg-org/ezmsg-sigproc/issues/42). When that's done that will be a nice feature add over the current state.

Already in my fledgling refactor to ezmsg.sigproc, I was able to add some new features:

* slicer can slice based on the CoordinateAxis data (e.g., `selection="Ch0, Ch3, Ch5"`)
* ranged_aggregate can aggregate over irregular axes
* bandpower and cwt don't have to pretend like the frequency axis is linear

Adding the `.astype(int)` to `index` and making a decision on whether to override `__eq__` are the only 2 issues I'd like to see resolved before merging. The other suggestions for `index` aren't critical.

griffinmilsap:
@cboulay Really great work on a coords refactor!  Just saw the project roadmap too, I'll start hopping in more there.

First off, `.astype(int)` is a great idea, I added it to `index`.

With respect to `__eq__`, I did some more research into the right way to handle equivalence here; it turns out this is a [known problem](https://stackoverflow.com/questions/51743827/how-to-compare-equality-of-dataclasses-holding-numpy-ndarray-boola-b-raises) with `dataclasses`.  A suggestion in that thread is to handle this with an 'array-safe' comparison, manually.. but this looks very inefficient for a check that might be run thousands of times per second. 

 I handled it by manually implementing the equivalence check using the [cpython implementation] (https://github.com/python/cpython/blob/e5e70c5f094c92f72789dd85ab4d74516a5b4932/Lib/dataclasses.py#L1108) as inspiration.  It should be noted that if a class defines its own custom `__eq__`, the `@dataclass` decorator won't create an `__eq__` method for the class, __but if you wrap a new class with `@dataclass` and the __base-class__ defines its own custom `__eq__`, the decorator will over-write the base-class `__eq__`__ which makes this issue a little tricky to solve in the general case.  Classes that derive from `AxisArray` or `ArrayWithNamedDims` (including `CoordinateAxis`) should probably be decorated with `@dataclass(eq=False)` to preserve the base-class `__eq__` method, if comparison is important to you.

cboulay (approved):

griffinmilsap:
@cboulay @pperanich Performance wise, this amount of dataclass inheritance has a bit of an impact.  I ran these tests on:
* Macbook Pro M1 2021
* `Python 3.11.10 (main, Sep  9 2024, 20:58:40) [Clang 18.1.8 ] on darwin`

__Command__: `uv run python -m timeit -n 500000 -s "from ezmsg.util.messages.axisarray import AxisArray; import numpy as np" "AxisArray(np.arange(20), dims = ['time'], axes = {'time': AxisArray.Axis.TimeAxis(fs = 100)})"`
* `dev`: 500000 loops, best of 5: __738 nsec per loop__
* `feature/add_coords_attrs`: 500000 loops, best of 5: __1.33 usec per loop__

Although its now twice as slow to create an AxisArray, I'm not sure this change is impactful relative to the general system performance.

```
(dev) ezmsg $ python -m ezmsg.util.perf_test
2024-11-11 14:27:54.817 - pid: 33507 - MainThread - INFO - test_performance: Running load test for dynamic size: 8 bytes
2024-11-11 14:27:54.833 - pid: 33507 - MainThread - INFO - run: Running in 2 processes.
2024-11-11 14:27:55.021 - pid: 33509 - TaskThread - INFO - log_result: Load test subscriber started. (PID: 33509)
2024-11-11 14:27:55.021 - pid: 33507 - TaskThread - INFO - publish: Load test publisher started. (PID: 33507)
2024-11-11 14:27:56.738 - pid: 33507 - TaskThread - WARNING - broadcast: SYSTEM/PUBLISHER/OUTPUT under subscriber backpressure!
2024-11-11 14:27:57.021 - pid: 33507 - TaskThread - INFO - publish: Exiting publish
2024-11-11 14:27:57.021 - pid: 33507 - TaskThread - INFO - wrapped_task: SYSTEM/PUBLISHER:publish Complete
2024-11-11 14:27:57.027 - pid: 33509 - TaskThread - INFO - log_result: Samples received: 39591
2024-11-11 14:27:57.028 - pid: 33509 - TaskThread - INFO - log_result: Sample rate: 19796.054570296288 Hz
2024-11-11 14:27:57.028 - pid: 33509 - TaskThread - INFO - log_result: Mean latency: 0.00013366225741890024 s
2024-11-11 14:27:57.028 - pid: 33509 - TaskThread - INFO - log_result: Total latency: 5.29182243347168 s
2024-11-11 14:27:57.028 - pid: 33509 - TaskThread - INFO - log_result: Data rate: 0.1583684365623703 MB/s
2024-11-11 14:27:57.028 - pid: 33509 - TaskThread - INFO - log_result: Dropped samples: 0 (0.0%)
2024-11-11 14:27:57.028 - pid: 33509 - TaskThread - INFO - wrapped_task: Normal Termination raised in SYSTEM/SUBSCRIBER:log_result
2024-11-11 14:27:57.028 - pid: 33507 - TaskThread - INFO - shutdown: Samples sent: 39591
2024-11-11 14:27:57.047 - pid: 33507 - MainThread - INFO - run: All processes exited normally
```

```
(feature/add_coords_attrs) ezmsg $ python -m ezmsg.util.perf_test
2024-11-11 14:27:31.948 - pid: 33221 - MainThread - INFO - test_performance: Running load test for dynamic size: 8 bytes
2024-11-11 14:27:31.968 - pid: 33221 - MainThread - INFO - run: Running in 2 processes.
2024-11-11 14:27:32.091 - pid: 33223 - TaskThread - INFO - log_result: Load test subscriber started. (PID: 33223)
2024-11-11 14:27:32.091 - pid: 33221 - TaskThread - INFO - publish: Load test publisher started. (PID: 33221)
2024-11-11 14:27:33.823 - pid: 33221 - TaskThread - WARNING - broadcast: SYSTEM/PUBLISHER/OUTPUT under subscriber backpressure!
2024-11-11 14:27:34.091 - pid: 33221 - TaskThread - INFO - publish: Exiting publish
2024-11-11 14:27:34.091 - pid: 33221 - TaskThread - INFO - wrapped_task: SYSTEM/PUBLISHER:publish Complete
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - log_result: Samples received: 38041
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - log_result: Sample rate: 19022.22112772893 Hz
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - log_result: Mean latency: 0.0001431899768803975 s
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - log_result: Total latency: 5.447089910507202 s
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - log_result: Data rate: 0.15217776902183144 MB/s
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - log_result: Dropped samples: 0 (0.0%)
2024-11-11 14:27:34.097 - pid: 33223 - TaskThread - INFO - wrapped_task: Normal Termination raised in SYSTEM/SUBSCRIBER:log_result
2024-11-11 14:27:34.098 - pid: 33221 - TaskThread - INFO - shutdown: Samples sent: 38044
2024-11-11 14:27:34.118 - pid: 33221 - MainThread - INFO - run: All processes exited normally
```

cboulay:
Great work @griffinmilsap ! LGTM.

src/ezmsg/util/generator.py
```diff
@@ -56,7 +56,7 @@ def compose(*funcs):
 
 # State class to hold the generator instance
 class GenState(ez.State):
-    gen: Generator
```
griffinmilsap, line 59: Default Generator typehint is `Generator[Unknown, None, None]` which caused a lot of linting errors for me using pylance

griffinmilsap:
I just added a bunch of typehinting fixes and changed the tests over to not use the now deprecated AxisArray.Axis.TimeAxis now that I've validated that the tests pass.  I'm going to merge into dev and test out on my other packages

cboulay:
In much of my sigproc code, AxisArray is not created often, but I use `replace` very heavily.

`uv run python -m timeit -n 500000 -s "from dataclasses import replace; from ezmsg.util.messages.axisarray import AxisArray; import numpy as np; template=AxisArray(data=np.arange(20), dims=['time'], axes={'time': AxisArray.Axis.TimeAxis(fs = 100)}); new_data=np.arange(22) - 2" "replace(template, data=new_data, axes={'time': replace(template.axes['time'], offset=100 * template.axes['time'].gain)})"`

* `dev`: 500000 loops, best of 5: 1.25 usec per loop
* `feature/add_coords_attrs`: 500000 loops, best of 5: 1.98 usec per loop

@griffinmilsap, I'd like to quickly revisit [this conversation](https://github.com/ezmsg-org/ezmsg/pull/127#issuecomment-2162962676). Do we still want to use `replace` instead of creating new objects?

griffinmilsap:
> @griffinmilsap, I'd like to quickly revisit [this conversation](https://github.com/ezmsg-org/ezmsg/pull/127#issuecomment-2162962676). Do we still want to use `replace` instead of creating new objects?

The reason to support `replace` is that we canonically support users making subclasses of `AxisArray`. This was more of a need prior to this feature addition.  I think there is much less of a need now that we can add arbitrary metadata and coordinates to `AxisArray` objects.

I did a search through all my ezmsg projects and found every instance of `(AxisArray):`.  I found 8 different instances where I was subclassing explicitly to create coordinate labels (things like `ch_names`, `class_names`, etc) and one instance where I did so in order to associate some metadata with my `AxisArray`.  At least in all of _my_ use cases, my code could all be converted to use existing mechanisms in the new `AxisArray` to store this new data.  

I'm amenable to _not_ requiring `replace` everywhere, but my remaining concerns are:
1. As mentioned, I already extensively sub-class AxisArray and it'll be `some` work to fix that (really not too much, but I can't speak for others here)
2. There might still be valid reasons to subclass AxisArray that I'm not already thinking about.  Off the top of my head, I could see users potentially wanting to add convenience properties/methods to a custom message type that's a subclass of AxisArray that expose properly typed metadata while enforcing particular naming conventions in their `attrs` and `coords`

If the potential speedup is substantial and the effort for refactoring existing code is manageable, I'd still want to evaluate what performance gain we get for losing the ability to subclass AxisArray.

griffinmilsap:
Looks like dataclass inheritance is actually QUITE expensive.  I addressed some of the typehinting issues using a protocol and duplicated some implementation details to get rid of all of the subclassing and recovered a lot of performance.

---------
AxisArray Creation.  __NOTE__ the change from `AxisArray.Axis.TimeAxis` (the old deprecated code that throws a warning) to `AxisArray.TimeAxis`
``` bash
(dev) ezmsg ツ uv run python -m timeit -n 500000 -s "from ezmsg.util.messages.axisarray import AxisArray; import numpy as np" "AxisArray(np.arange(20), dims = ['time'], axes = {'time': AxisArray.TimeAxis(fs = 100)})"
500000 loops, best of 5: 778 nsec per loop
```

``` bash
(fix/axisarray_optim) ezmsg ツ uv run python -m timeit -n 500000 -s "from ezmsg.util.messages.axisarray import AxisArray; import numpy as np" "AxisArray(np.arange(20), dims = ['time'], axes = {'time': AxisArray.TimeAxis(fs = 100)})"
500000 loops, best of 5: 541 nsec per loop
```
----------
`replace` benchmark

``` bash
(dev) ezmsg ツ uv run python -m timeit -n 500000 -s "from dataclasses import replace; from ezmsg.util.messages.axisarray import AxisArray; import numpy as np; template=AxisArray(data=np.arange(20), dims=['time'], axes={'time': AxisArray.TimeAxis(fs = 100)}); new_data=np.arange(22) - 2" "replace(template, data=new_data, axes={'time': replace(template.axes['time'], offset=100 * template.axes['time'].gain)})"
500000 loops, best of 5: 1.21 usec per loop
```

``` bash
(fix/axisarray_optim) ezmsg ツ uv run python -m timeit -n 500000 -s "from dataclasses import replace; from ezmsg.util.messages.axisarray import AxisArray; import numpy as np; template=AxisArray(data=np.arange(20), dims=['time'], axes={'time': AxisArray.TimeAxis(fs = 100)}); new_data=np.arange(22) - 2" "replace(template, data=new_data, axes={'time': replace(template.axes['time'], offset=100 * template.axes['time'].gain)})"
500000 loops, best of 5: 974 nsec per loop
```

I'll make a new PR for this change.

PR #155, opened 2024-11-12
Create and use fast_replace

cboulay:


griffinmilsap:
Well, the performance doesn't lie; this is a pretty significant speedup on an operation we do a LOT.

Would it be possible to move replace to `messages.util` (or even `messages.__init__`) (instead of AxisArray), and make it possible to use a monkeypatch or possibly even an environment variable for all code to fall back to using original (slow) `replace`.  Unless the idea is that replace is only to be used on AxisArrays, in which case it can remain in AxisArray and should be typehinted appropriately (with the TypeVar bound on AxisArray)

Additionally, the original [`dataclasses.py:replace`](https://github.com/python/cpython/blob/865f096f51bb7290eb10b33b330f387d245c2472/Lib/dataclasses.py#L1580) has a check to deliver a helpful error message if its used on a non-dataclass object; I'll defer to you if that's a useful add here or not.  

Once thats addressed, I'm happy to take this for a spin in a new tagged release on PyPi (with the new AxisArray w/ coords :D)

PR #156, opened 2024-11-12
Add functionality to AxisArray to convert to XArray DataArray

laurendiaz:


griffinmilsap:
I've got a few changes I'd like to push into this one and add a test before I merge.  @cboulay are you alright with adding xarray as a test dependency?  It seems like it slows down tests slightly on first import

cboulay:
ezmsg-org/ezmsg-sigproc unit tests are about 100x longer than ezmsg-org/ezmsg unit tests. I'm not concerned about the added CI runner time. (BTW, we went over our free alotment last month because I made a bunch of small ezmsg-sigproc changes).

I guess my only concern with adding this as a test dependency is that it might prevent us from testing against Python 3.13 earlier than we might otherwise want to. I'm sure there's a way we could tell pyproject.toml to ignore xarray deps on python 3.13.

griffinmilsap:
I *think* I made the xarray test fully optional and specified the install of xarray as a test dependency if and only if we're on a python version below 3.13.  I can't seem to test this yet because I can't get `uv sync --all-extras --dev --python 3.13` or even 3.12 to execute due to an issue with `distutils being removed from Python 3.12`..? Its weird because the github runners don't seem to choke on it..  idk what the issue is on my side.

griffinmilsap (approved):
I made some changes and added a test.  I think this is good to go!

PR #159, opened 2024-11-12
Dev -> Main

griffinmilsap:
Lots of AxisArray enhancements in this merge!


griffinmilsap:
@cboulay @pperanich I bumped the version and noticed that it was duplicated; once in `src/ezmsg/core/__init__.py` and once in `pyproject.toml`.  I just modified `src/ezmsg/core/__init__.py` to the following are we alright with this?

``` python
import importlib.metadata

__version__ = importlib.metadata.version("ezmsg")
...
```

@cboulay  I know you've done some really cool vcs versioning with `ezmsg-sigproc` but afaik that's not standard practice yet.  Cool to move to this?

pperanich:
If you ran `uv sync` it should have automatically updated the version string in the `__init__.py` file from the version specified in the `pyproject.toml`.

griffinmilsap:
Ahhh that wasn't clear to me.  It just looked like it was redundantly specified.  Are there pros to that approach vs this solution?  Happy to revert that

pperanich (approved):

PR #161, opened 2024-11-26
Move numpy dep from optional to required

### LINKED ISSUE #160: numpy optional for AxisArray
I guess this rarely comes up because almost everything using ezmsg also uses numpy, but not everything! (ezmsg-zmq for example)
This should be a quick fix. I'll go straight to main.

cboulay:
Fixes #160

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,12 +14,14 @@ classifiers = [
   "Operating System :: OS Independent",
 ]
 requires-python = ">=3.8"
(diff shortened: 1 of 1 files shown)


cboulay:
@griffinmilsap or @pperanich , this should be the easiest "code review" of your lives.

While I'm willing to bypass the ->dev-> gitflow step for this, I don't want to bypass the review-required step unless it's an emergency. This is urgent, but not emergency.

pperanich (approved):

PR #163, opened 2024-11-28
Move numpy dep from required to extras - AxisArray and test

### LINKED ISSUE #160: numpy optional for AxisArray

cboulay:
Closes #160 
* Moves numpy dep from required to extras - AxisArray and test
* Adds helpful error message when user attempts to import from `.axisarray` and numpy is not installed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "numpy>=1.24.4",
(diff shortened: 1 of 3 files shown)


cboulay:
bump

griffinmilsap (approved):
LGTM!

PR #169, opened 2025-02-19
Merge fixes to "Mermaid" graph visualization and MessageLogger

griffinmilsap:
Last merge to main was in November, and since then there's been lots of work to fix the half-implemented "Mermaid" graph visualization functionality, and there has also been a performance-oriented fix/feature for MessageLogger.


cboulay (approved):
I did the mermaid graph changes and Preston approved those so maybe we can assume they don't need to be reviewed again.

I just looked at the message logger and it looks good to me. I would like if there were a unit test that the buffer clears properly on shutdown but I won't hold up the PR for that.

griffinmilsap:
@cboulay thanks for the approval; I was about to write that unit test until I noticed there's .. absolutely no unit tests for MessageLogger.  I'll need to add those in a future commit.

I'm going to merge this now because it addresses a performance issue that's been a pressing issue for upcoming deliverables, but I'll make a note to add unit tests for MessageLogger in the near future.

PR #177, opened 2025-04-22
feat: first pass at removing requirement to inherit directly from ez.…

pperanich:
This is a first pass at removing the requirement that a Unit's `SETTINGS` and `STATE` inherit directly from `ez.Settings` and `ez.State`, respectively. Instead, for `SETTINGS`, we ensure the type is a frozen dataclass, and for `STATE`, we ensure the type is a dataclass.


griffinmilsap:
@pperanich and I chatted about this PR; it might make sense to redo parts of this PR to not even require dataclasses for state/settings.  Some feedback we got from other users indicate that people would like to use Pydantic or Param data structures that enforce bounds on settings, for example.  I would like to add additional checks/features if users do choose to use dataclasses, but not require it if they want to try their luck with other data structures.

griffinmilsap:
This is pretty stale.  We'll come back to this after the new backend changes percolate in.

Issue #178, opened 2025-05-03
What is the recommended way to have multiple simultaneous non-interacting pipelines?


cboulay:
Let's say I have these two contrived pipelines.

Pipeline A:
gamepad capture -> temporal filtering -> quantization -> thresholding -> event triggers -> publish to logger

Pipeline B:
brain signals -> temporal filtering -> downsample -> spatial filtering -> spectrogram -> adaptive scaler -> publish to visualizer

* These pipelines must run simultaneously and continuously (for hours).
* There will never be a need for the graphs to pass messages back and forth.
* I must be able to interact with the graphs when they are running
    * using `ezmsg mermaid`
    * attaching-detaching new nodes transiently to spy on the data

Idea 1: Start 2 independent graph services with 2 calls to `ezmsg serve --address 127.0.0.1:{26000,26100}`, then pass one of the addresses each to the `ez.run(..., graph_address=("127.0.0.1", {26000, 26100}))` invocations for the pipelines.
 
* `ezmsg serve` always attempts to start the SHMServer and always at the same address (`127.0.0.1:25979`) so the second call to `ezmsg serve` fails.

Idea 2: Don't pre-start the graph services; just run the pipelines with `ez.run(..., graph_address=None)`
 
* The GraphService port will be the first free port counting up from 10000 -- non-deterministic because it depends on what else is running and the order in which I run pipelines A and B.
* We could prepend each invocation with `EZMSG_SERVER_PORT_START=26x00`, where `x` is unique to each pipeline. This remains non-deterministic but less likely to have collisions, though zombie pipelines are a concern.
* Note: We cannot specify the `graph_address` if the `GraphService` is not already running because specifying the `graph_address` implies it should exist otherwise the call to `ez.run` will fail.

Idea 3: Run the following Python code:

```Python
shm_address_str = os.environ.get(
    SHMSERVER_ADDR_ENV, f"127.0.0.1:{SHMSERVER_PORT_DEFAULT}"
)
shm_address = Address.from_string(shm_address_str)
shm_service = SHMService(shm_address)
shm_server = shm_service.create_server()
graphs = []
for gs_ix in range(2):
    graph_service = GraphService(Address("127.0.0.1", 26000 + 100 * gs_ix))
    graph_server = graph_service.create_server()
    graphs.append((graph_service, graph_server))
# snipped cleanup code
```
Then use the known addresses in `ez.run`.

* Ports are deterministic and zombie pipeline risk is mitigated.
[...]

PR #186, opened 2025-07-14
Merge GraphServer and SHMServer

griffinmilsap:
Addresses #178 

This commit refactors the ezmsg core backend to merge the GraphServer and the SHMServer.  Rationale in #178; There is no SHMServer anymore, and the GraphServer manages SHM leases.  

Originally, SHMServer was separated from GraphServer because I had a mind to support distributed systems.  Now that GraphServer is coupled to SHM blocks, distributed systems that connect to a remote graph server will necessarily need to communicate via TCP, even if they're on the same machine.  In practice, I'm not sure this ever worked; and a more well-thought-out implementation is probably worth thinking through.  

This is a pretty significant refactor of the backend, so we might want to test this a bit more proactively.

### DIFF
--- a/src/ezmsg/core/backend.py
+++ b/src/ezmsg/core/backend.py
@@ -17,7 +17,6 @@ from .stream import Stream
 from .unit import Unit, PROCESS_ATTR
 
 from .graphserver import GraphService
-from .shmserver import SHMService
 from .graphcontext import GraphContext
 from .backendprocess import (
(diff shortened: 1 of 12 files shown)


cboulay:
All of my ezmsg-sigproc tests pass. Konrad told me he went through the PR in detail and is happy with it. I'll merge this in.

PR #189, opened 2025-08-13
Fix: GraphServer port no-longer lingers

griffinmilsap:
When (re)launching a GraphServer on the canonical port immediately after stopping the server, python throws `OSError: Address already in use`.  Setting `SO_REUSEADDR` addresses this issue.  

Additionally, we manually set the socket nonblocking which is required for asyncio. I think it actually does this internally when we pass it to `start_server` but at least we're explicit here.

Finally, we are much more explicit about server shutdown and we don't swallow exceptions anymore.

### DIFF
--- a/src/ezmsg/core/netprotocol.py
+++ b/src/ezmsg/core/netprotocol.py
@@ -177,22 +177,29 @@ def create_socket(
     ignore_ports: typing.List[int] = RESERVED_PORTS,
 ) -> socket.socket:
     sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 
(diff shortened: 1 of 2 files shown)


KonradPilch:
So, apparently, my comments on the Server merge PR were not visible because I forgot to click 'Submit Review'. My comment is even more applicable to this PR I believe:

Basically, `test_attach` is skipped with the reason provided being "canonical port isn't always available". Previously, if not skipped and the default port was available, after running it would leave the default port lingering for me in (presumably) `TIME_WAIT` mode.  This PR fixes this! <- I have checked this. 

It still has the issue of failing when the address is being used. I suggest amending line 111 to:
```python
    try:
        graph_server = graph_service.create_server()
    except OSError:
        pytest.skip("Default GraphServer port not available")
        return
```

I can also make this change and push it if you like.

Issue #188, opened 2025-08-08
Regression: Comms via TCP


griffinmilsap:
The core ezmsg communication model supports transmission of data via a local `MessageCache`, SHM transport, and TCP in that order of fallback.  

If a `Publisher` and a `Subscriber` are...
1. In the same process, they just use the `MessageCache` and `TX_LOCAL` to transmit data
2. In different processes on the same machine, they use SHM transport and `TX_SHM` messages
3. On different machines (or `force_tcp`) is set, the whole message is serialized then transmitted with `TX_TCP`

For this logic to work, pubs/subs need a mechanism to determine which process and machine they're on.  Process ID is easy enough, but getting a unique identifier for a machine is a little less trivial.  A quick Google search revealed that the `node` argument within a UUID is usually derived from the machine's ethernet MAC address, and several StackOverflow posts recommend `uuid.get_node()` for getting a machine identifier.

While this approach worked for a while, MacOS recently decided that ethernet MAC addresses should be hidden from userspace for privacy concerns, and the [behavior of the function is unpredictable across various versions of Python.](https://github.com/astral-sh/python-build-standalone/issues/587).  Interestingly, `get_node()` mostly just returns random numbers on most of our tested platforms, including linux machines, which is not a problem when using `fork` multiprocessing because the same random number is assigned -- but on MacOS, which uses `spawn` multiprocessing by default, we get different random numbers, which results in .. you guessed it -- all messages are transferred using TCP.

Systems that rely on multiprocessing within ezmsg have been facilitating message transport using TCP for a little while now, which has resulted in a regression of the perf test.  TCP telemetry in ezmsg is already quite poorly optimized #5 , and this doesn't help.

Proposed fix:
Instead of trying to logic our way through this, let's attempt local and SHM transfer of a random number on connect and bypass the whole issue of finding a reliable machine identifier.

cboulay:
Closed via #190

PR #190, opened 2025-08-13
Fix: TCP Comms Regression

griffinmilsap:
Addresses #188 

This is a minimal-diff fix for the mentioned issue, but there's some additional optimization that I'd like to implement with a more substantial backend comms rework that I've charted out.  For now though, we could merge this fix and benefit from SHM transfer via multiprocessing.

### DIFF
--- a/src/ezmsg/core/netprotocol.py
+++ b/src/ezmsg/core/netprotocol.py
@@ -87,7 +87,8 @@ class PublisherInfo(ClientInfo):
 
 
 @dataclass
-class SubscriberInfo(ClientInfo): ...
+class SubscriberInfo(ClientInfo):
+    shm_access: bool = False
 
 
 def uint64_to_bytes(i: int) -> bytes:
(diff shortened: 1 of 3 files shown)


cboulay:
I retargeted this on `fix/server-merge`. If we merge _that_ PR and delete the branch then this PR will automatically retarget itself on `dev`.

cboulay:
Anecdotally...

* `python -m ezmsg.util.perf_test --many-dynamic-sizes`
    * (M4Pro, Python 3.12) message throughput did not change appreciably
    * (M4Pro, Python 3.11) message throughput did not change appreciably
* ezmsg-sigproc unit test suite (excluding integration tests)
    * (M4Pro, Python 3.12) runtime went from 1m 23s to 1m 20s
    * (M4Pro, Python 3.11) runtime went from 1m 22s to 1m 21s

I'm sure any differences are within the margin of error. I wasn't expecting any changes in the unit test suite because the ezmsg-sigproc tests don't really leverage shared memory. I ran them mostly to prove that they still work and there were no performance regressions.
lgtm!

PR #192, opened 2025-08-15
Bump Python deps; python>=3.10.15; arrange dev deps in groups

### LINKED ISSUE #157: Remove python 3.8 from tests

cboulay:
Small refactor of pyproject.toml with big impact

* Python dependency now >= 3.10.15
* `uv sync` will now install dev dependencies
* dev dependencies now in groups, not optional (extras)
* I kept `docs` dependencies as an optional (`uv sync --extra docs`), not a group, because it lists ezmsg-sigproc and I don't want that installing every time we do `uv sync` on `ezmsg`.

closes #157

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   pytest:
     strategy:
       matrix:
-        python-version: ["3.10.15", "3.11", "3.12", "3.13"]
+        python-version: ["3.10.15", "3.11"]  # Omit until shutdown fixed: , "3.12", "3.13"]
         os:
           - "ubuntu-latest"
           - "windows-latest"


KonradPilch:
I like the change!
To confirm the intended usage, `uv sync` will install all dev dependencies, but to get all dependencies (numpy for `AxisArray` and all docs dependencies), we should use `uv sync --all-extras`?

cboulay:
> we should use `uv sync --all-extras`?

Correct.

Remember that `uv sync` is probably only relevant to this repo when used by developers, so in practice numpy comes anyway with the normal dev group dependencies thus we don't need that extra when using `uv sync`. In other words, you don't technically need `--all-extras`, you could do `--extra docs`. Side note, I'm not sure the docs deps should go in `extras`, maybe they should also be in a `group`, but I didn't want to change that part because I know you're working on it.

IMO, `extras` are best suited for dependencies that we want non-developers to have the option of installing when using `pip`: `pip install "path/to/ezmsg[axisarray]" or even `pip install "ezmsg[axisarray]"` directly from pypi.

You and I have zero use cases where we don't already have other dependencies that require numpy so we've never had to add this extra to our `ezmsg` entry in the dependency list, but if someone didn't have another transitive dependency on numpy AND they wanted to use AxisArray then they would need to add this.

No one needs to `pip install "ezmsg[docs]"` which is why I don't think that group belongs in `extras`.

cboulay:
OK fixed. If you want the docs dependencies, it's now `uv sync --group docs` or `uv sync --all-groups`

cboulay:
>You mentioned in https://github.com/ezmsg-org/ezmsg/issues/157 that using init=True in dataclasses (available in Python 3.10) may break fast_replace? Do you know if this is true?

No, I don't remember the details around that. I'm happy to proceed and deal with it when it arrives.

PR #193, opened 2025-08-15
AxisArray.concatenate now concatenates along CoordinateAxis

cboulay:


tests/messages/test_axisarray.py
```diff
@@ -234,35 +273,47 @@ def test_sliding_win_oneaxis(nwin: int, axis: int, step: int):
     assert np.array_equal(res, expected)
     assert np.shares_memory(res, expected)
 
+
 def xarray_available():
     try:
         import xarray
-        return True 
+
```
KonradPilch, line 280: Is this a ruff formatting change?  Seems a little awkward to have the return expression from the try branch "attached" to the except statement.

KonradPilch:
I think this is a useful feature. I am happy with the PR. We may want to additionally add a tiny test for the failure modes (ie. `dim`s don't match or specified `dim` doesn't exist in all `AxisArray`s).

Issue #195, opened 2025-09-08
Process Unit adds ~40ms delay each


Lechros:
When I run the test code below, each Process `ez.Unit` adds about 40ms delay.

If I run the same logic with manual `multiprocessing.Process`, the delay is only about 2ms.

Also, I see that in the first or second iteration the delay is normal, but after that it always grows to ~40ms. Maybe this is related to backpressure or buffering?

Environment:
- Ubuntu 22.04
- Python 3.10.15 and 3.12 (same result)

Test code:
```python
import asyncio
from dataclasses import dataclass
from datetime import datetime
from typing import AsyncGenerator, Tuple

import ezmsg.core as ez


@dataclass
class TestData:
    seq: int
    at: datetime


class TestSettings(ez.Settings):
    max_seq: int


class Producer(ez.Unit):
    SETTINGS = TestSettings

    OUTPUT = ez.OutputStream(TestData)

    @ez.publisher(OUTPUT)
    async def publish(self) -> AsyncGenerator:
        for seq in range(1, self.SETTINGS.max_seq + 1):
            await asyncio.sleep(0.1)
            now = datetime.now()
            print(f"{seq}: start time: {now}")
            yield self.OUTPUT, TestData(seq, now)


class Step(ez.Unit):
    INPUT = ez.InputStream(TestData)
    OUTPUT = ez.OutputStream(TestData)

    @ez.subscriber(INPUT, zero_copy=True)
    @ez.publisher(OUTPUT)
    async def delay(self, msg: TestData) -> AsyncGenerator:
        yield self.OUTPUT, msg


class Consumer(ez.Unit):
    INPUT = ez.InputStream(TestData)

    @ez.subscriber(INPUT, zero_copy=True)
    async def subscribe(self, msg: TestData) -> None:
        now = datetime.now()
        print(f"{msg.seq}: end time: {now}, took {(now - msg.at).total_seconds() * 1000:.1f}ms")


class System(ez.Collection):
    PRODUCER = Producer()
    STEP1 = Step()
    STEP2 = Step()
    STEP3 = Step()
    STEP4 = Step()
    STEP5 = Step()
    CONSUMER = Consumer()

    def configure(self) -> None:
        settings = TestSettings(max_seq=10)

        self.PRODUCER.apply_settings(settings)

    def network(self) -> ez.NetworkDefinition:
        return (
            (self.PRODUCER.OUTPUT, self.STEP1.INPUT),
[...]

PR #196, opened 2025-09-17
HOTFIX: Address Linux latency regression

griffinmilsap:
Addresses #195 by disabling Nagle's algorithm on ezmsg's TCP sockets.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ezmsg"
-version = "3.6.1"
+version = "3.6.2"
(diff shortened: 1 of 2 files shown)


cboulay:
Good to go.
Note: This repo is the one ezmsg repo where versioning isn't automatic based on git tags. Would you like to bump the pypproject.toml version in this PR or to do that in `main` before doing a release?

griffinmilsap:
Ah good call with the version bump.  I'll do that in this PR.  I'll also open a new PR to merge this fix into dev.

PR #198, opened 2025-09-17
Backend Rework: Message Channels

griffinmilsap:
WIP


griffinmilsap:
Okay, I think this is ready for merge into dev.  @cboulay @KonradPilch Feel free to do so at your leisure.

tests/ez_test_utils.py
```diff
@@ -20,14 +19,15 @@ def get_test_fn(test_name: str | None = None, extension: str = "txt") -> Path:
...
-
-    return file_path
+    # Create a unique temporary file name to avoid collisions when running the
+    # full test suite in parallel or when other tests use the same test name.
+    # Use NamedTemporaryFile with delete=False so callers can open/remove it.
+    prefix = f"{test_name}-" if test_name else "test-"
+    tmp = tempfile.NamedTemporaryFile(
+        prefix=prefix, suffix=f".{extension}", delete=False
+    )
+    tmp.close()
```
KonradPilch: Should we also make sure to delete these temporary files?

src/ezmsg/core/shm.py
```diff
@@ -215,32 +206,45 @@ def size(self) -> int:
...
+
     Tracks the SharedMemory object and manages client connection leases.
     When all leases are released, the shared memory is automatically cleaned up.
     """
+
     shm: SharedMemory
     leases: set["asyncio.Task[None]"] = field(default_factory=set)
 
+    @classmethod
+    def create(cls, num_buffers: int, buf_size: int) -> "SHMInfo":
```
KonradPilch: When we drop dependency on 3.10, can replace the return value here with `typing.Self`. It will avoid typing issues if we ever (unlikely though) subclass `SHMInfo`.

src/ezmsg/core/messagecache.py
```diff
@@ -1,133 +1,123 @@
...
-class CacheMiss(Exception): 
-    """
-    Exception raised when a requested message is not found in cache.
-    
-    This occurs when trying to retrieve a message that has been evicted
-    from the cache or was never stored in the first place.
-    """
-    ...
 
+class MessageCache:
```
KonradPilch: Can probably adapt the docstring we had for `Cache` before.

src/ezmsg/core/messagecache.py
```diff
@@ -1,133 +1,123 @@
...
-        :rtype: Generator[Any, None, None]
-        :raises CacheMiss: If message not found in cache or shared memory.
+        Reconstitute a message in mem and keep it in cache, releasing and
+        overwriting the existing slot in cache.
+        This method passes the lifecycle of the memoryview to the MessageCache
+        and the memoryview will be properly released by the cache with `free`
+
+        :param mem: Source memoryview containing serialized object.
+        :type from_mem: memoryview
+        :raises UninitializedMemory: If mem buffer is not properly initialized.
```
KonradPilch: Can also raise `CacheMiss` on release of the memory (in call to `_release` via `_put`)

src/ezmsg/core/messagecache.py
```diff
@@ -1,133 +1,123 @@
...
+                mem.release()
+
+    def release(self, msg_id: int) -> None:
         """
-        Clear all cached messages and identifiers.
-        
-        Resets all cache slots to None, effectively clearing the entire cache.
+        Release memory for the entry associated with msg_id
+
+        :param mem: Source memoryview containing serialized object.
```
KonradPilch: This looks like the docstring for `put_from_mem`

src/ezmsg/core/subclient.py
```diff
@@ -68,22 +66,36 @@ async def create(
...
 
     def __init__(
-        self, id: UUID, topic: str, graph_service: GraphService, **kwargs
+        self, id: UUID, topic: str, graph_address: AddressType | None, **kwargs
     ) -> None:
         """
         Initialize a Subscriber instance.
-        
+
+        DO NOT USE this constructor, use Subscriber.create instead.
```
KonradPilch: As with the Publisher, maybe we enforce this?

src/ezmsg/core/subclient.py
```diff
@@ -149,151 +149,75 @@ async def _graph_connection(
...
+                    # While this currently works, its non-obvious what this
+                    # accomplishes and why its implemented this way.  I just
+                    # wasted 2 hours removing this seemingly un-necessary event
+                    # to introduce a bug that is only resolved by reintroducing
+                    # the event.  Some thought should be put into replacing the
+                    # bespoke communication protocol with something a bit more
+                    # standard (JSON RPC?) with a more common/logical pattern
+                    # for creating/handling comms.  We will probably want to
+                    # keep the bespoke comms for the publisher->channel link
+                    # as that is in the hot path.
```
KonradPilch: I agree with the comment, but definitely something for a different PR, especially if it's only a case of readability.

KonradPilch (commented):
First, AMAZING work. Thanks @griffinmilsap.

Just leaving some comments, most are minor. 

Very happy with the changes. If you think this is in a state you're happy with @griffinmilsap, I'd be ok with merging once you have looked at my comments and decided what does and doesn't need actioning.

KonradPilch:
>  GitHub Pages docs covering the low-level Publisher/Subscriber API
> GitHub Pages docs covering message transport with sequence diagrams

Also @griffinmilsap, feel free to add an explanation of how the backend works now in the [docs repo](https://github.com/ezmsg-org/ezmsg-org.github.io/tree/main/docs). I feel like the explanation you've provided in this PR is exactly what would work best there. If you want to move onto the list of further work you've got listed above, let me know and I can put the docs work on my todo instead.

src/ezmsg/core/graphcontext.py
```diff
@@ -33,95 +34,103 @@ class GraphContext:
...
         :type from_topic: str
         :param to_topic: The destination topic name
         :type to_topic: str
         """
-        await self._graph_service.disconnect(from_topic, to_topic)
+        await GraphService(self.graph_address).disconnect(from_topic, to_topic)
         self._edges.discard((from_topic, to_topic))
 
-    async def sync(self, timeout: float | None = None) -> None:
+    async def sync(self, timeout: typing.Optional[float] = None) -> None:
```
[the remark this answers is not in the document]
griffinmilsap: Thats my bad; I had a hell of a merge that I had to manage conflicts for after the docstring update came through.  I just missed this one.  I'll update that.

src/ezmsg/core/backendprocess.py
```diff
@@ -306,8 +305,8 @@ async def shutdown_units() -> None:
 
             asyncio.run_coroutine_threadsafe(shutdown_units(), loop=loop).result()
 
-            for cache in MessageCache.values():
-                cache.clear()
+            # for cache in MessageCache.values():
+            #     cache.clear()
```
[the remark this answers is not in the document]
griffinmilsap: This was a lazy hack to clean up the cached memory segments prior to deallocating all SHMs, but now we have proper cleanup with ChannelManager deleting unused Channels (each of which clear their own message caches.  Also, MessageCache used to be a global object but now its a local cache for each MessageChannel.

src/ezmsg/core/messagemarshal.py
```diff
@@ -71,43 +74,45 @@ def to_mem(cls, msg_id: int, obj: Any, mem: memoryview) -> None:
...
-        :type mem: memoryview
-        :return: Message ID if memory is initialized, None otherwise.
-        :rtype: int | None
+        Get msg_id from a buffer; if uninitialized, return None.
+
+        :param mem: buffer to read from.
+        :type mem: memoryview | bytes
+        :return: Message ID of encoded message
+        :rtype: int
+        :raises UndersizedMemory: If buffer is not initialized.
```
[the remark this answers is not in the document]
griffinmilsap: Good catch; fixed.

src/ezmsg/core/messagecache.py
```diff
@@ -1,133 +1,123 @@
...
-from uuid import UUID
-from contextlib import contextmanager
+from .messagemarshal import MessageMarshal
 
-from .shm import SHMContext
-from .messagemarshal import MessageMarshal, UninitializedMemory
 
-from collections.abc import Generator
-from typing import Any
+class CacheMiss(Exception): ...
```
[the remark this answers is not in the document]
griffinmilsap: Oops, its back now.

src/ezmsg/core/messagecache.py
griffinmilsap: Adapted; it's close enough :)

src/ezmsg/core/messagecache.py
griffinmilsap: I think it cannot raise `CacheMiss` on `_release`; there's no call to `self.__getitem__`; instead we directly access `self._cache.__getitem__` which won't raise `CacheMiss`

src/ezmsg/core/messagecache.py
griffinmilsap: Oops. Fixed.

src/ezmsg/core/subclient.py
griffinmilsap: Enforced.

tests/test_new_messaging.py
```diff
@@ -0,0 +1,129 @@
+import asyncio
```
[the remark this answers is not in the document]
griffinmilsap: I forgot about this file! it was intended as an example for how to use the low level API.  Its a bit complicated, but demonstrates multiprocess communications with Publisher Subscriber and how to manually start a GraphServer.  I've cleaned it up and put it in `examples`.

tests/test_perf_configs.py
```diff
@@ -0,0 +1,99 @@
+import contextlib
```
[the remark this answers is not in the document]
griffinmilsap: I'm going to leave this as is for now, but consider yourself heard.

src/ezmsg/core/shm.py
KonradPilch: I agree. This shows up in a few places, so it'll be a nice little update in the future.

tests/test_reconstitute.py
```diff
@@ -0,0 +1,131 @@
+import asyncio
```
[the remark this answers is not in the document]
KonradPilch: Definitely looked like an interesting experiment.

tests/ez_test_utils.py
```diff
@@ -23,11 +27,11 @@ def get_test_fn(test_name: str | None = None, extension: str = "txt") -> Path:
...
-    tmp = tempfile.NamedTemporaryFile(
-        prefix=prefix, suffix=f".{extension}", delete=False
-    )
-    tmp.close()
-    return Path(tmp.name)
+    tmp = tempfile.NamedTemporaryFile(prefix=prefix, suffix=f".{extension}")
+    try:
+        yield Path(tmp.name)
+    finally:
+        tmp.close()
```
KonradPilch: This works on Linux and MacOS, because they allow opening a file that has an active handler, but Windows does not, so you'll notice that the tests are failing on the Windows runner. 
Here is a quick fix:
```
    tmp = tempfile.NamedTemporaryFile(prefix=prefix, suffix=f".{extension}")
    tmp.close()  # Close so others can open it on Windows
    path = Path(tmp.name)
    try:
        yield path
    finally:
        try:
            path.unlink()
        except FileNotFoundError:
            pass
```

KonradPilch: A completely alternative solution is to not create our own temporary files at all - pytest provides a fixture (`tmp_path`) for that. See [here](https://docs.pytest.org/en/stable/how-to/tmp_path.html). So, the tests could look like:
```
def test_hello(tmp_path):
    test_filename = tmp_path / "test_hello.txt"
    ...
    with open(test_filename, "r") as file:
        ...
```
A little fiddlier perhaps and we'd need to adjust retention number parameters, but worth considering.

tests/ez_test_utils.py
KonradPilch: @griffinmilsap this solution breaks on Windows, see [this comment](https://github.com/ezmsg-org/ezmsg/pull/198/files/9f6f94a2e143b9f23ea5c64c010477d8c4ea8265..438b557bc012c0b8cf77348ea181d6b245e460f4#r2558905502).

tests/ez_test_utils.py
griffinmilsap: Its always windows...  ugh.

Applied your suggested fix, tests seem to be passing now.

src/ezmsg/core/messagecache.py
```diff
@@ -1,133 +1,123 @@
...
 
-class Cache:
-    """
-    Shared-memory backed cache for objects.
-    
-    Provides a buffer cache that can store objects both in memory
-    and in shared memory buffers, enabling efficient message passing between
-    processes with automatic eviction based on buffer age.
-    """
+    def __init__(self, num_buffers: int) -> None:
```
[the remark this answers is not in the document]
griffinmilsap: aye; I added an assert in MessageChannel.py.

PR #203, opened 2025-10-28
Transition from RTD to GitHub Pages docs deployment

cboulay:


cboulay:
It took a bit longer to get this ready for merging because we refactored. There's now a repo ezmsg-org.github.io that has the narrative docs. Then each of our namespace packages can have their own API-focused docs, which is now what we have in this repo.
They can all be found at www.ezmsg.org

PR #206, opened 2025-12-05
Feature: Profiling Hooks

griffinmilsap:
## Profiling Feature Overview
- Adds sliding-window profiling for publishers and channels to surface throughput, byte rate, and backpressure/processing latency without extra serialization hops.
- GraphServer can now fan out `PROFILE` requests over existing control links and aggregate JSON snapshots across all connected publishers/channels.

## Key Details
- Telemetry primitives in `src/ezmsg/core/profiling.py` track rolling counts/bytes and buffer lease durations; `Backpressure` optionally emits timings per client.
- Publishers (`src/ezmsg/core/pubclient.py`) record msg/byte rates and backpressure lease timing; channels (`src/ezmsg/core/messagechannel.py`) capture per-subscriber processing time and last transport mode (`local|shm|tcp`), plus subscriber handles for readability.
- GraphServer (`src/ezmsg/core/graphserver.py`) understands new protocol commands `PROFILE`/`PROFILE_DATA` (`src/ezmsg/core/netprotocol.py`), gathers profiles, and exposes them via `GraphService.profile(window_s)`.
- CLI support: `python -m ezmsg.core profile --window 2.0` prints aggregated JSON, including per-publisher/channel stats.
- Subscribers may pass `handle=` when creating to label profiles; defaults to the topic/address.

## Why It Matters
- Gives per-process, per-stream visibility into bottlenecks: time from backpressure lease→free (upper bound on publish rate) and per-subscriber processing latency inside `ez.subscriber` callbacks.
- Minimal overhead via small time windows (~2s default) and reuse of existing graph connections; no extra message serialization.

## Usage Example
```bash
# Collect a 2s snapshot from a running graph
ezmsg profile --window 2.0

# Programmatic
from ezmsg.core.graphserver import GraphService
profile = await GraphService().profile(2.0)
```

## Notable Outputs
- Publisher: total/window messages & bytes, message/byte rates, backpressure avg/max (ms), in-flight leases.
- Channel: per-subscriber avg/max processing (ms), samples, message rate, in-flight leases, transport mode, subscriber handle/id.

### DIFF
--- a/tests/test_channelmanager.py
+++ b/tests/test_channelmanager.py
@@ -20,7 +20,7 @@ class DummyChannel:
         self.waited = False
         self.local_bp: dict = {}
 
-    def register_client(self, client_id, queue, local_backpressure):
+    def register_client(self, client_id, queue, local_backpressure, handle=None):
         self.clients[client_id] = queue
         if local_backpressure is not None:
             self.local_bp[client_id] = local_backpressure


griffinmilsap:
Discovered limitations with this approach that are handled better in #234

PR #210, opened 2026-01-30
Deprecate ezmsg.util.generator

cboulay:
AI coding assistants will sometimes grab on to these old patterns and try to introduce them into new code. I'd like to make their deprecation more explicit to avoid that from happening. I've already done so for several of the namespace packages like ezmsg-sigproc (not released, but in `dev`).


cboulay:
There are still a few usages of @consumer that we might want to deprecate.
* `modify_axis`
* `set_key`

cboulay:
OK, no longer warning, mostly docstring changes.

PR #211, opened 2026-02-07
Add Array API support to sliding_win_oneaxis

cboulay:
There's a very small performance hit when in numpy (a couple hasattr checks), but moderate overall performance gains in other backends (70%), especially as the chunk sizes grow.

I think I can improve upon this further. Previously, I was relying on the fact that manipulating the strides creates a simple view on the data, so I could make a 1-sample-step array then simply index it with `::step_size` with no data copies. 

However, with non-strided array APIs, the 1-sample-step array is expensive. I think the non-strided code path should skip the 1-sample-step intermediate and copy only the data required in the output.

Converting this to draft while I work on that...

### DIFF
--- a/src/ezmsg/util/messages/axisarray.py
+++ b/src/ezmsg/util/messages/axisarray.py
@@ -2,7 +2,7 @@ from abc import abstractmethod, ABC
 from collections.abc import Callable, Generator, Iterable
 from contextlib import contextmanager
 from dataclasses import field, dataclass
-from array_api_compat import get_namespace, is_numpy_array, is_torch_array
+from array_api_compat import get_namespace, is_cupy_array, is_numpy_array, is_torch_array
 import math
 import typing
 import warnings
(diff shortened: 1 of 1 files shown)


tests/messages/test_axisarray.py
```diff
@@ -275,6 +275,107 @@ def test_sliding_win_oneaxis(nwin: int, axis: int, step: int):
...
+    if nwin == 0:
+        assert res.size == 0
+        return
+
+    # Compare against the strided numpy result.
+    expected = sliding_win_oneaxis(data, nwin, axis, step)
+    np.testing.assert_array_equal(res, expected)
+
+
+@pytest.mark.benchmark(group="sliding_win")
```
griffinmilsap: dang I didn't know you could do benchmarks like this.  Wondering if this is the pattern to follow for the perf test suite.  how do benchmarks/perf tests work on github's CI runners (which I assume aren't performance machines?)

griffinmilsap (approved):
LGTM; even with tests!

tests/messages/test_axisarray.py
cboulay: I manually triggered a tests run on this branch and they do print the output:

https://github.com/ezmsg-org/ezmsg/actions/runs/21844998402

I think it's trivial to disable pytest-benchmark and I think that should be our default.

Issue #13, opened 2023-01-17
Request: Tests for graceful shutdown and force termination on KeyboardInterrupt


griffinmilsap:
This behavior has been broken on Windows for at least one minor version, and is now fixed in 3.1..  I do most my dev on MacOS, but Windows is a significantly different platform and doesn't handle signals (and a lot of things) quite the same way as *nix systems.

I'd like a test that launches an ezmsg system that's supposed to run for two seconds, publishing some preset number of messages (20 @ 10 msgs/sec?) to a receiver that logs the messages to a file using `MessageLogger` before terminating with `NormalTermination`.  The test should then interrupt the system with a KeyboardInterrupt (__how???__) after one second, and verifies that shutdown has happened gracefully before all expected messages were received/logged by the receiver.  The system should shutdown without raising any exceptions to pass the test.  It's important to write the test so that we cover "single process mode" where only one process will be executed (without multiprocessing) and a system where the publisher and receiver are both executed in different processes (because interrupts and graceful shutdown/termination are handled differently in this mode of execution).  

Finally, a separate system should be written that tests the "force terminate" functionality that occurs when KeyboardInterrupt is sent again after graceful shutdown has occurred.  This could be tested by blocking keyboard interrupt with the `signal` module by using the following line in the `initialize` function of a `ez.Unit`: `signal.signal(signal.SIGINT, signal.SIG_IGN)`.  The unit should have a single task that calls a blocking `time.sleep` for ~10 seconds.  Again, after a second the test harness should inject a KeyboardInterrupt, and after another second, the test harness should inject a second Keyboard interrupt.  Its okay for the system to raise exceptions when terminated this way; and the test passes if the system shuts down before 10 seconds elapses (ideally around 1-2 seconds of execution).  This should be done with one unit (and one process) as well as multiple units/multi-process mode.

griffinmilsap:
Also all tests should be implemented across fork and spawn multiprocessing!

PR #220, opened 2026-02-24
Fix: Multiple shutdown issues and enhance shutdown UX

### LINKED ISSUE #219: On shutdown, task was destroyed but it is pending
When shutting down an ezmsg session (sandboxed graph service) with Ctrl+C, I'm getting a wall of error messages:
- Task was destroyed but it is pending! for Publisher, Subscriber, SHM, and Channel tasks
- RuntimeError: Event loop is closed in `close_stream_writer` when `writer.close()` is called after the loop has shut down
- RuntimeError: coroutine ignored GeneratorExit for publisher serve/channel coroutines

Root cause: During shutdown, `DefaultBackendProcess.process()` calls `context.revert()` which properly cancels and awaits publisher/subscriber tasks. However, the event loop (`new_threaded_event_loop`) is then stopped and closed without cancelling/awaiting any remaining orphaned tasks. These orphaned tasks include:
- `complete_tasks` (subscriber handlers, unit task coroutines) that were cancelled by `monitor_termination` but never awaited
- Channel tasks, SHM monitor tasks, and other background tasks spawned transitively

Additionally, `close_stream_writer()` in netprotocol.py doesn't handle the case where `writer.close()` raises `RuntimeError` because the event loop is already closed.

cboulay:
Fixes #219 

src/ezmsg/core/netprotocol.py — Wrapped writer.close() in a try/except RuntimeError so that if the event loop is already closed when close_stream_writer is called, we return early instead of raising RuntimeError: Event loop is closed.

src/ezmsg/core/backendprocess.py — Added a _cancel_remaining() coroutine in the finally block of new_threaded_event_loop that collects all pending tasks, cancels them, and awaits their completion (with a 5-second timeout) before stopping the loop. This ensures orphaned tasks (subscriber handlers, channel tasks, SHM monitors, etc.) are properly cleaned up instead of being destroyed with a pending warning.

### DIFF
--- a/tests/test_clean_shutdown.py
+++ b/tests/test_clean_shutdown.py
@@ -6,6 +6,8 @@ import threading
 import time
 from pathlib import Path
 
+import pytest
+
 
 ROOT = Path(__file__).resolve().parents[1]
 RUNNER = Path(__file__).with_name("shutdown_runner.py")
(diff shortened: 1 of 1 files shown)


griffinmilsap:
And at this point I've added tests to evaluate SIGINT and forced SIGINT (double-tap) shutdown behavior with a bunch of misbehaving systems as well as well-behaving ones!  This has been a long-time coming (closes #13)

PR #248, opened 2026-05-19
Feature: Log to file for `GraphServer` CLI

### LINKED ISSUE #246: Add log-to-file support for `ezmsg start` (background GraphServer)
## Problem
`ezmsg start` spawns the GraphServer as a background subprocess, but the child's stdout/stderr aren't captured anywhere. If the server misbehaves, there's no log to consult after the fact.

The alternative, `ezmsg serve`, runs in the foreground and prints logs to the console, but then the user has to keep that terminal open. That's at odds with the use case of the graph service as a long-running, service-like process that should be able to live in the background.

## Proposed solution
Give the GraphServer a default log file location and let users override it.

**Default path (platform-appropriate):**
- Windows: `%LOCALAPPDATA%\ezmsg\logs\GraphServer-{port}.log`
- macOS/Linux: `~/.local/share/ezmsg/logs/GraphServer-{port}.log` (or the equivalent `$XDG_DATA_HOME` location)

The `{port}` suffix keeps logs from multiple concurrent GraphServers separate.

**Override mechanism:**
- An environment variable (e.g. `EZMSG_LOG_FILE` or `EZMSG_LOG_DIR`)
- A CLI argument on `ezmsg start` / `ezmsg serve` (e.g. `--log-file`)

CLI argument should take precedence over the env var, which takes precedence over the default.

## Scope
This should be a small change, roughly:
- Add a file handler to the ezmsg logger when running the GraphServer, pointed at the resolved log path (creating the parent directory if needed).
- Plumb a `--log-file` argument through the relevant `ezmsg` subcommands and pass it to the spawned subprocess in `start`.
- Document the default location and override options.

Log rotation, structured logging, and similar niceties are out of scope for this initial change; a single file per port is enough to unblock the "run it as a service" use case.

griffinmilsap:
Closes #246 

## Summary

Add file logging for the `ezmsg start` / `ezmsg serve` service envelope.

This change adds `--log-file` support to both commands, passes the option through from `start` to the spawned `serve` subprocess, and otherwise falls back to `EZMSG_LOG_FILE` or a default per-run log path under a platform-appropriate ezmsg data directory. Default logs are organized by GraphServer port and use a timestamped filename for each execution.

## Scope

This intentionally stays narrow and only affects the `start` / `serve` workflow. It does **not** attempt to define broader logging behavior yet for `ez.run`, `GraphRunner`, subprocess/session aggregation, or remote clients.

## Testing

- Added one parser test for `serve --log-file`
- Added one command test verifying `start` passes `--log-file` through to `serve`
- Ran:

```bash
uv run pytest tests/test_command.py -q
```

### DIFF
--- a/src/ezmsg/core/commands/common.py
+++ b/src/ezmsg/core/commands/common.py
@@ -1,4 +1,8 @@
 import argparse
+import logging
+import os
+from datetime import datetime
+from pathlib import Path
 
 from ..graphserver import GraphService
 from ..netprotocol import Address
@@ -6,6 +10,14 @@ from ..netprotocol import Address
 
 def add_address_argument(parser: argparse.ArgumentParser) -> None:
     parser.add_argument("--address", help="Address for GraphServer", default=None)
+
+
+def add_log_file_argument(parser: argparse.ArgumentParser) -> None:
+    parser.add_argument(
+        "--log-file",
+        help="Path to the ezmsg service log file",
+        default=None,
+    )
 
 
 def add_compact_argument(parser: argparse.ArgumentParser) -> None:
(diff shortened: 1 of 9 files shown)


src/ezmsg/core/commands/common.py
```diff
@@ -8,6 +12,14 @@ def add_address_argument(parser: argparse.ArgumentParser) -> None:
     parser.add_argument("--address", help="Address for GraphServer", default=None)
 
 
+def add_log_file_argument(parser: argparse.ArgumentParser) -> None:
```
KonradPilch, line 18: Need to plumb this into `run_command()` too.

src/ezmsg/core/commands/common.py
```diff
@@ -24,3 +36,51 @@ def graph_address_from_args(args: argparse.Namespace) -> Address:
...
+    log_path = log_file.expanduser().resolve()
+    log_path.parent.mkdir(parents=True, exist_ok=True)
+
+    logger = logging.getLogger("ezmsg")
+    if not any(
+        isinstance(handler, logging.FileHandler)
+        and getattr(handler, "baseFilename", None) == str(log_path)
+        for handler in logger.handlers
+    ):
+        formatter = next(
```
KonradPilch: We can just define the standard ezmsg formatter and use that.

src/ezmsg/core/commands/common.py
```diff
@@ -24,3 +36,51 @@ def graph_address_from_args(args: argparse.Namespace) -> Address:
...
+            (
+                handler.formatter
+                for handler in logger.handlers
+                if handler.formatter is not None
+            ),
+            None,
+        )
+        handler = logging.FileHandler(log_path, encoding="utf-8")
```
KonradPilch: we should probably deal with lifecycle management here, so that we don't leave open handlers we created.

KonradPilch (approved):
I like what you've done. I have a few suggestions which I have added as commits to this branch. Please look through them. My changes based on my review:

1. Wired `log_file` into `run_command` to avoid an `AttributeError`.
2.  Lifecycle management of logging handlers we create. 
3. Defined a standard logging formatter in `logconfig.py` that is accessed via `create_ezmsg_log_formatter()`.
4. Split logs between stdout and stderr based on the log level (for when other processes own the Graph Service).

cboulay (approved):

